PR #20, opened 2018-05-07
Do not check for multiscale if path is dataset

hanslovsky:
`N5HDF5Reader.list( String path )` throws unchecked exception if `path` is dataset. Therefore, check if `path` is a dataset before listing contents.


hanslovsky:
@axtimwalde This will probably solve your exception issue.

PR #98, opened 2018-06-20
Remove segmentFrom from segmentToFragmentsMap before manipulating the…

hanslovsky:
… fragments for segmentTo

 - handles the case that segmentFrom == merge.segmentId
 - not 100% sure why such a merge was generated. Could be related to (resolved) bug that actions would be saved repeatedly
 - should also deal with segmentInto == segmentFrom == merge.segmentId
   - get fragments for segmentFrom (remove from map)
   - add new fragment list {into} for segment segmentInto
   - add fragments for segmentFrom to {into}


hanslovsky:
@axtimwalde I was able to load your projects with the changes in this PR. Please check out dc4e7ef and see if the fragment-segment assignments look sane to you. I will merge once you confirm.

hanslovsky:
Was confirmed in person.

PR #109, opened 2018-06-28
Label multiset type argmax

### LINKED ISSUE #102: Require label to block list mapping instead of creating it on the fly
Block lookup is too slow and there is no reason this cannot be pre-computed.

### LINKED ISSUE #107: Consider switching back to IntegerType only for labels
1. `LabelMultisetType` does not compress well and restricting to two or three entries per pixel minimze benefits
 2. considerable overhead with special case implementations, e.g. #106 
 3. plans to require label id to block list mapping makes unique list of contained label ids in `VolatileLabelMultisetTypeArray` obsolete

hanslovsky:
Fixes #107 
Note that meshes are now only possible if a paintera dataset with the appropriate sub-directory `label-to-block-mapping` exists. Currently, this is stored in single files per id, which is a lot of tiny files. In the future, think about a database instead. Also, I haven't figured out how to update block lists when a block does not contain a label anymore (the label was completely overpainted for that block).

Important note: All existing label multiset datasets need to be re-generated.

Update: Also fixes #102


hanslovsky:
This seems to be working now (tested on my laptop with cremi), going to merge and release 0.1.0

PR #126, opened 2018-08-20
Undo for assignment actions

### LINKED ISSUE #125: Add undo for assignment actions
Accidental merge of two segments cannot be undone via single detach action (if segments are big, this will require a lot of work!). Undo would be very helpful for such scenarios.

hanslovsky:
Fixes #125 

Need to use a little more to make sure everything works as intended before merge.


hanslovsky:
Successfully tested:
 - saving to Paintera project
 - committing LUT

PR #133, opened 2018-09-06
Add label utilities dependency

hanslovsky:
Use [label-utilities](https://github.com/saalfeldlab/label-utilities) for label-block-lookup with n5 backend. This solves the issue of many small files on nrs and makes it possible to use paintera for bigger data sets.


hanslovsky:
Will need to test a little before merging.

hanslovsky:
Tested and confirmed:
 - start empty project
  - add n5 data sets
  - create new label dataset
 - load lauritzen project with synapses
 - run `PainteraShowContainer` on training snapshot

will merge now
cc @axtimwalde

PR #136, opened 2018-09-12
N5 imglib2 bounded

hanslovsky:
Major refactor: Improve cache management
 - use global cache instead of one cache per source
 - upper bound on memory or entries (still to decide that).


hanslovsky:
Merge this for now, might need some updates in the future., e.g. the clean-up of the backing cache

PR #149, opened 2018-10-24
Priority executor service

hanslovsky:
Use a priority queue to render data to screen instead of a plain `ExecutorService`. That way, tiles of interest can be rendered first, and more important render jobs will run sooner, even if submitted later. This will eliminate the forced delay for painting after navigation actions or similar.
See also the [priority-executor-service-experimental(https://github.com/saalfeldlab/paintera/tree/priority-executor-service-experimental) branch.

This is WIP and not yet ready to merge but this is to track process.


hanslovsky:
As we plan to go back to full screen rendering for navigation and only keep the tiled rendering for painting, I will close this PR. I will not delete the relevant branch(es) as they might contain useful code for future implementations of such rendering mechanism.

PR #173, opened 2018-12-07
Build fixes: Eclipse Oxygen, Oracle JDK 10

igorpisarev:
This PR fixes the build in the latest version of Eclipse Oxygen & Oracle JDK 10. The reason is likely a compiler bug because the code was certainly valid but the implicit conversions seemed to trigger false positive compiler errors.

### DIFF
--- a/src/main/java/org/janelia/saalfeldlab/fx/MenuFromHandlers.java
+++ b/src/main/java/org/janelia/saalfeldlab/fx/MenuFromHandlers.java
@@ -144,29 +144,4 @@ public class MenuFromHandlers {
 		}
 
 	}
-
-	public static void main(String[] args)
-	{
-		PlatformImpl.startup(() -> {});
-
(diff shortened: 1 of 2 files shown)


src/main/java/org/janelia/saalfeldlab/fx/MenuFromHandlers.java
```diff
@@ -149,12 +149,11 @@ public static void main(String[] args)
 	{
 		PlatformImpl.startup(() -> {});
 
-		ContextMenu menu = new MenuFromHandlers(Arrays.asList(
```
hanslovsky: The `main` method is obsolete (was used to test functionality). Please remove the entire method in this PR.

igorpisarev: Done

PR #180, opened 2019-01-11
Render tiles with padding to fix interpolation artifacts in 3d orthoslice view

igorpisarev:
Fixes gaps and interpolation artifacts between the meshes forming the 3D orthoslice view.
* 1px padding is applied to each rendered tile on each side (the padding is always the same regardless of the screen scale used)
* Instead of using the grid layout to arrange tile images, they are now cropped (to remove the padding) and drawn onto a canvas at their corresponding grid position
* Texture coordinates for the meshes are recomputed each time the rendered tile image changes

Here is how it looked before the fix:
mesh_gaps

Before the fix but with adjusted mesh size:
mesh_interpolation

After the fix:
mesh_fixed

### DIFF
--- a/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
+++ b/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
@@ -577,6 +577,9 @@ public class ViewerPanelFX
 
 		final CellGrid cellGrid = grid.getGrid();
 		final long[] gridPos = new long[2];
+		final int[] fullCellDims = new int[2];
+		cellGrid.cellDimensions(fullCellDims);
+
 		for (int i = 0; i < grid.numTiles(); ++i) {
 			final long[] cellMin = new long[2];
 			final int[] cellDims = new int[2];
@@ -587,7 +590,7 @@ public class ViewerPanelFX
 			image.addListener((obs, oldv, newv) -> {
 				if (newv == null) {
 					canvasPane.getCanvas().getGraphicsContext2D().setFill(Color.BLACK);
-					canvasPane.getCanvas().getGraphicsContext2D().fillRect(cellMin[0], cellMin[1], cellDims[0], cellDims[1]);
+					canvasPane.getCanvas().getGraphicsContext2D().fillRect(cellMin[0], cellMin[1], fullCellDims[0], fullCellDims[1]);
 				} else {
 					final int[] padding = imageDisplayGrid.get().getPadding();
 					canvasPane.getCanvas().getGraphicsContext2D().drawImage(
@@ -595,7 +598,7 @@ public class ViewerPanelFX
 						padding[0], padding[1], // src X, Y
 						newv.getWidth() - 2 * padding[0], newv.getHeight() - 2 * padding[1], // src width, height
 						cellMin[0], cellMin[1], // dst X, Y
-						cellDims[0], cellDims[1] // dst width, height
+						fullCellDims[0], fullCellDims[1] // dst width, height
 					);
 				}
 			});
--- a/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
+++ b/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
@@ -48,6 +48,7 @@ import javafx.collections.ObservableList;
 import javafx.scene.Node;
 import javafx.scene.image.Image;
 import javafx.scene.layout.StackPane;
+import javafx.scene.paint.Color;
 import net.imglib2.Point;
 import net.imglib2.Positionable;
 import net.imglib2.RealLocalizable;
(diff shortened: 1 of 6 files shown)


hanslovsky:
Great stuff, thanks @igorpisarev 
I will look into it in detail this week

igorpisarev:
I pushed one more commit with a better fix for the visualization problem with border tiles.
The issue was with scaling factors that were computed based on the tile size, so they ended up being different for regular tiles and border tiles. To fix this, I added full tile dimensions as a parameter to the renderer. Then, the rendered tile image is upscaled and drawn onto the displayed canvas, also using the full tile dimensions to ensure that the scaling factors are consistent.

igorpisarev:
I tested and pushed a proper fix for interpolation artifacts and scaling issues with the border tiles.

To illustrate the problem that I've been attempting to fix: let's say the screen scale is set to 0.01, so that 1px of rendered data corresponds to 100px on the screen. When a rendered border image was drawn on the canvas, it was upscaled to the dimensions of the displayed border tile, so if it's smaller than 100px in this example, the data looked squashed.
To properly upscale rendered tiles, I added the screen scale parameter to be passed along with each rendered image. This issue is now fixed for both 2D view and 3D orthoslice view.

I also put a lower bound on the screen scale depending on the renderer block size in `RenderUnit`. This is to prevent excessive upscaling of the rendered tile images when arranging them on the canvas so they would start overlapping with each other.

hanslovsky:
Thanks @igorpisarev!
I looked at some data, setting rather extreme screen scales (down to 0.001) and it looks good now. I will still need to have a look at the code before merging.

src/main/java/bdv/fx/viewer/render/RenderUnit.java
```diff
@@ -339,7 +346,16 @@ public void paint() {
 					null
 			);
 
-			renderTarget.drawOverlays(image::set);
+			if (renderedScreenScaleIndex != -1)
+			{
+				final Double renderedScreenScale;
+				synchronized (RenderUnit.this)
+				{
+					renderedScreenScale = renderedScreenScaleIndex >= 0 && renderedScreenScaleIndex < screenScales.length ? screenScales[renderedScreenScaleIndex] : null;
```
hanslovsky: Is the `>= 0` check necessary? We already checked `renderedScreenScaleIndex != -1`

src/main/java/bdv/fx/viewer/render/RenderUnit.java
```diff
@@ -339,7 +346,16 @@ public void paint() {
 					null
 			);
 
-			renderTarget.drawOverlays(image::set);
+			if (renderedScreenScaleIndex != -1)
+			{
+				final Double renderedScreenScale;
+				synchronized (RenderUnit.this)
```
hanslovsky: Why synchronize on `RenderUnit.this` here? Is it becasue `screenScales` may change during the call to `screenScales[screenScaleindex]`?

src/main/java/bdv/fx/viewer/ViewerPanelFX.java
```diff
@@ -570,32 +570,49 @@ public OverlayPane getDisplay()
...
-			final ImagePane imagePane = new ImagePane(cellDims[0], cellDims[1]);
-			imagePane.imageProperty().bind(image);
-			LOG.debug("Putting render block {} into cell at position {}", i, gridPos);
-			pane.add(imagePane, (int) gridPos[0], (int) gridPos[1]);
+
+			final int[] imageSize = new int[2], unpaddedImageSize = new int[2];
+			final ReadOnlyObjectProperty<RenderUnit.RenderedImage> renderedImageProperty = grid.renderedImagePropertyAt(i);
+			renderedImageProperty.addListener((obs, oldv, newv) -> {
+				if (newv == null) {
+					canvasPane.getCanvas().getGraphicsContext2D().setFill(Color.BLACK);
```
hanslovsky: Please add comment that `newv == null` check and filling with black color is only necessary because there is no `src` `BlendMode` in JavaFX, also describe the `src` blend mode like this:
> Overwrites the destination with the source, regardless of alpha

Paste these links into the comment as well:
https://bugs.openjdk.java.net/browse/JDK-8092156
https://books.google.com/books?id=UxM2iXiFqbMC&pg=PA97&lpg=PA97&dq=Porter-Duff+%22source+over+destination%22&source=bl&ots=hYj5U2X5Lk&sig=ACfU3U095X67T4FghqCpgYrAhEWbNtd_xQ&hl=en&sa=X&ved=2ahUKEwjyg4rRufXfAhXvuFkKHfK8BysQ6AEwAnoECAcQAQ#v=snippet&q=%22overwrites%20the%20destination%20with%20the%20source%2C%20regardless%20of%20alpha%22&f=false
https://docs.oracle.com/javase/8/javafx/api/javafx/scene/effect/BlendMode.html

igorpisarev:
As I have found, the current revision has an issue with the scaling of the border tiles between different screen scale levels, so they slightly jump when the rendered image changes from one screen scale to another. I have not been able to figure out the reason for this, and the code has already become slightly overcomplicated for the initial problem I aimed to solve.

I decided to use a different (simpler) approach and extend the border tiles to be the same size as regular tiles. To avoid changelog clutter, I created a revised version #195 and am going to close this one.

PR #195, opened 2019-01-30
Fix interpolation artifacts between render tiles [revised]

igorpisarev:
This is a revised version of #180. The difference is that the issue with scaling of border tiles here is solved differently: instead of rendering only the truncated part and keeping track of scaling coefficients, the border tiles are rendered in full, and then only the visible part is used.

Summary of changes:

- 1px padding is applied to each rendered tile on each side (the padding is always the same regardless of the screen scale used)
- Instead of using the grid layout to arrange tile images, they are now cropped (to remove the padding) and drawn onto a canvas at their corresponding grid position
- Texture coordinates for the meshes are recomputed each time the rendered tile image changes
- Border tiles are extended to be of the same size as regular tiles
- After a tile is rendered, its alpha channel is set to 1 to avoid blending problems

### DIFF
--- a/src/main/java/bdv/fx/viewer/render/TransformAwareBufferedImageOverlayRendererFX.java
+++ b/src/main/java/bdv/fx/viewer/render/TransformAwareBufferedImageOverlayRendererFX.java
@@ -35,6 +35,7 @@ import java.util.function.Consumer;
 
 import javafx.scene.image.Image;
 import net.imglib2.realtransform.AffineTransform3D;
+import net.imglib2.type.numeric.ARGBType;
 import net.imglib2.ui.TransformListener;
 import org.janelia.saalfeldlab.fx.util.InvokeOnJavaFXApplicationThread;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


src/main/java/bdv/fx/viewer/render/TransformAwareBufferedImageOverlayRendererFX.java
```diff
@@ -99,6 +100,23 @@ public void drawOverlays(final Consumer<Image> g)
...
+				 * See discussion about possibility of adding this mode in the future versions and more information here:
+				 *
+				 * https://bugs.openjdk.java.net/browse/JDK-8092156
+				 *
+				 * https://books.google.com/books?id=UxM2iXiFqbMC&pg=PA97&lpg=PA97&dq=Porter-Duff+%22source+over+destination%22&source=bl&ots=hYj5U2X5Lk&sig=ACfU3U095X67T4FghqCpgYrAhEWbNtd_xQ&hl=en&sa=X&ved=2ahUKEwjyg4rRufXfAhXvuFkKHfK8BysQ6AEwAnoECAcQAQ#v=snippet&q=%22overwrites%20the%20destination%20with%20the%20source%2C%20regardless%20of%20alpha%22&f=false
+				 *
+				 * https://docs.oracle.com/javase/8/javafx/api/javafx/scene/effect/BlendMode.html
+				 */
+				for (final ARGBType px : sourceImage.asArrayImg())
+					px.set(px.get() | (0xff << 24));
```
hanslovsky: Does the compiler optimize this? I would prefer to have a static final integer for the alpha value, e.g.
```java
private static final int FULL_OPACITY = 0xff << 24;
```

igorpisarev: Pretty sure constant expressions are optimized, `ARGBType` internals also look the same.
I can add it to better conform to the project's coding style.

hanslovsky:
Thank you @igorpisarev

PR #199, opened 2019-02-05
Serialize initial lut supplier in paintera project

### LINKED ISSUE #198: Fragment-segment assignment is not applied when opening project second time
I have created a dataset in paintera format with a pre-defined `fragment-segment-assignment`, like:
```
root.n5/
  labels/
    data/
    fragment-segment-assignment/
    ...
```
When I create a new paintera project via
```
paintera [...] project_folder
```
and open the labels, the fragment segment assignment is applied correctly.

However, once I save the project, close paintera and reload it, the assignment is not applied anymore and I only see individual fragments.

hanslovsky:
Otherwise, initial lut cannot be loaded when reading label source from project file

Fixes #198

### DIFF
--- a/src/main/java/org/janelia/saalfeldlab/paintera/control/assignment/FragmentSegmentAssignmentOnlyLocal.java
+++ b/src/main/java/org/janelia/saalfeldlab/paintera/control/assignment/FragmentSegmentAssignmentOnlyLocal.java
@@ -6,6 +6,7 @@ import java.util.Optional;
 import java.util.function.LongSupplier;
 import java.util.function.Supplier;
 
+import com.google.gson.annotations.Expose;
 import gnu.trove.impl.Constants;
 import gnu.trove.iterator.TLongLongIterator;
 import gnu.trove.map.TLongLongMap;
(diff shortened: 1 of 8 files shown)


hanslovsky:
@constantinpape please  test on this PR. You will have to re-add the label sources for this to take effect (if not, the label sources would still load but no lookup would be applied, as you observed).

I will test for a while before merging, I want to make sure that everything works as expected.

Other users please test as well. Silence will be interpreted as confirmation.

constantinpape:
I checked out this PR and for me everything works as expected now:
after saving + reopening and commiting + reopening the correct fragment assignment is loaded.

constantinpape:
Ok, thanks!
What's your plan for the next release?
I would like to get this to collaborators and rather not build from source for this.

hanslovsky:
I cut releases rather frequently, so you should expect a release as soon as this is merged. I just want to make sure that everything works as expected with this fix, first.

hanslovsky:
I just opened one of the lauritzen blocks (probably the most complex projects that we have, currently) with a serialized fragment-segment assignment lookup. It ignores the lookup on paintera version 0.9.6 but it loads it with this patch. I will merge and release.

PR #200, opened 2019-02-05
Different rendering modes for navigation/painting

igorpisarev:
**Motivation**: multi-tile rendering (currently used rendering scheme) enables very responsive painting which is great. But, when browsing through a dataset, the tiles are rendered and updated at different times, and the picture may look inconsistent, especially during rapid zooming/translation/rotation.

This PR allows the renderer to work in two different modes: rendering the entire screen as a single image (for navigation), and rendering it as multiple tile images (for painting). This way, the picture looks consistent when browsing the data while painting is kept responsive.

------------------------------

**Changes:** The new class `RenderingModeController` switches between the modes when needed by using `renderUnit.setBlockSize()` and keeps track of some extra information.
When switching between the modes, repaint is generally not required because the canvas already displays the most recent state. However, when painting is initiated before the most recent viewer transformation has been rendered, the repaint is needed (otherwise, only the user-painted tile will have up-to-date transformation while the rest of the canvas will be left with an older frame).

To ensure this consistency, I added a tag to each rendered image that is incremented whenever the viewer transformation changes, or painting is started/finished. `RenderingModeController` also keeps track of the last received tag, so that it knows whether the most recent transformation has been displayed or not.

To avoid race conditions, the renderer code synchronizes on `viewerState` when fetching the current viewer transform and generating a tag for the new frame.

3D orthoslice views also need to be updated carefully when the mode switch occurs so that they always have a valid texture:
* *Single-tile -> multi-tile*: the current texture of the mesh is used to initialize the new smaller meshes.
* *Multi-tile -> single-tile*: the update is postponed until the next rendered frame is received (this way the texture will be available immediately)

------------------------------

**Demo**

Current master:

master-optimized

This PR:

pr-optimized

### DIFF
--- a/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
+++ b/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
@@ -394,7 +394,7 @@ public class ViewerPanelFX
 		state.setViewerTransform(transform);
 		for (final TransformListener<AffineTransform3D> l : transformListeners)
 			l.transformChanged(viewerTransform);
-		renderingModeController.transformChanged(transform);
+		renderingModeController.transformChanged();
 		requestRepaint();
 	}
 
@@ -556,6 +556,11 @@ public class ViewerPanelFX
 	public OverlayPane<?> getDisplay()
 	{
 		return this.overlayPane;
+	}
+
+	public RenderingModeController getRenderingModeController()
+	{
+		return renderingModeController;
 	}
 
 	/**
--- a/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
+++ b/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
@@ -32,7 +32,6 @@ package bdv.fx.viewer;
 import bdv.cache.CacheControl;
 import bdv.fx.viewer.render.RenderUnit;
 import bdv.fx.viewer.render.RenderingModeController;
-import bdv.fx.viewer.render.RenderingModeController.RenderingMode;
 import bdv.viewer.Interpolation;
 import bdv.viewer.RequestRepaint;
 import bdv.viewer.Source;
@@ -49,9 +48,7 @@ import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
 import javafx.scene.Node;
 import javafx.scene.canvas.Canvas;
-import javafx.scene.image.Image;
 import javafx.scene.layout.StackPane;
-import javafx.scene.paint.Color;
 import net.imglib2.Point;
 import net.imglib2.Positionable;
 import net.imglib2.RealLocalizable;
@@ -119,7 +116,7 @@ public class ViewerPanelFX
 
 	private final ObjectProperty<RenderUnit.ImagePropertyGrid> imageDisplayGrid = new SimpleObjectProperty<>(null);
 
-	private final RenderingModeController renderingModeController;
+	private RenderingModeController renderingModeController;
 
 	public ViewerPanelFX(
 			final List<SourceAndConverter<?>> sources,
@@ -228,7 +225,8 @@ public class ViewerPanelFX
 				cacheControl,
 				options.getTargetRenderNanos(),
 				options.getNumRenderingThreads(),
-				renderingExecutorService);
+				renderingExecutorService,
+				() -> renderingModeController.getCurrentTag());
 
 		this.renderingModeController = new RenderingModeController(renderUnit);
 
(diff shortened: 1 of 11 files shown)


src/main/java/bdv/fx/viewer/render/RenderingModeController.java
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	private static final int[] DEFAULT_TILE_SIZE = {250, 250};
+
+	private final RenderUnit renderUnit;
+	private RenderingMode mode;
+
+	private final int modeSwitchDelay = 250;
+	private final Timer modeSwitchTimer;
+	private TimerTask modeSwitchTimerTask;
```
igorpisarev: I ended up not using this `Timer` here at all. A different logic was implemented in https://github.com/saalfeldlab/paintera/pull/200/commits/7ced12a6cd9c9bf81ae0f29d8e9de463336145eb

Sorry for cluttering the changeset with intermediate changes that were removed in later commits. Maybe it's easier to review this PR in full, or group the first few commits as a single changeset: https://github.com/saalfeldlab/paintera/pull/200/files/7ced12a6cd9c9bf81ae0f29d8e9de463336145eb

src/main/java/bdv/fx/viewer/ViewerPanelFX.java
```diff
@@ -390,21 +390,24 @@ public void requestRepaint(final long[] min, final long[] max)
...
-	 * Get a copy of the current {@link ViewerState}.
+	 * Get the current {@link ViewerState}.
 	 *
-	 * @return a copy of the current {@link ViewerState}.
+	 * @return the current {@link ViewerState}.
 	 */
 	public ViewerState getState()
 	{
-		return state.copy();
+		return state;
```
igorpisarev: To be able synchronize on viewer state in `RenderUnit.Paintable.paint()`, I changed this method to return the actual object and not a copy. I also changed other places that used this method and relied on a copy of the state to use synchronization instead.

hanslovsky:
I just had a quick look and it works for me. I don't know if I will have time to go through the code this week, though.

One thing to consider is the issue with only partial updates to the highest screen scales when starting to paint before the single-tile renderer has finished painting the highest screen scales:
 1. set the highest screen scale such that rendering does not happen immediately
 2. scroll through data
 3. start paint immediately, before highest screen scale is rendered
 4. only the painted tiles get rendered to highest resolution (click on image for full res): paintera-screen-scales

A solution to that would be to switch to single-tile rendering whenever the drag is released in painting mode.

hanslovsky:
And another thought: do you think it would make sense to get rid of the told tiled renderer and instead pass a ROI to the multi resolution renderer and revert render Target? In that case, the executor service should probably not be a part of the renderer but be passed into the renderer with every call to render

igorpisarev:
I addressed the issue with partial screen updates when painting is started before the last frame is rendered at the highest resolution. With the fix, the screen will be re-rendered to the highest resolution after the painting is done if needed.

I like the idea of changing the renderer to operate on a single image but at different ROIs -- this way we would not have to re-create images and threads every time, and the code should become simpler.
(maybe that's how I should have initially approached this).
Another option is to optimize image and thread creation in the currently proposed version, but your suggestion is better.
I'll try to implement a quick prototype on a separate branch and will see if it's easier to merge it here or rebase and create a new PR instead.

hanslovsky:
> I like the idea of changing the renderer to operate on a single image but at different ROIs -- this way we would not have to re-create images and threads every time, and the code should become simpler.

I agree

> (maybe that's how I should have initially approached this).

Maybe, but it would also require more changes to the multiresolution renderer itself, e.g. tracking the bounding box for each repaint, which is probably complicated on its own.

> I'll try to implement a quick prototype on a separate branch and will see if it's easier to merge it here or rebase and create a new PR instead.

Great, thanks. I am OK to merge this in the meantime (haven't tested yet, though). Let me know what you prefer. I think it is OK to have all those efforts in the history so the different approaches (multi-resolution renderer, multi-tile renderer, combined, multi-resoluition renderer w/region of interested) are obvious in the git history.

igorpisarev:
I'm also fine with merging this PR so it's easier to keep track of what's been done.
Would really appreciate if you get a chance to briefly test and review the code. Thanks!

hanslovsky:
Just toying around with it and functionally it looks good to me. Will also look at the code!

src/main/java/bdv/fx/viewer/ViewerPanelFX.java
hanslovsky: While this comes at the risk of callers modifying the state directly, it is probably fine to not a return a copy here. Once (or if) we return to multi-resolution renderer w/region-of-interest rendering we should think about reverthing this change.

hanslovsky:
Code looks good to me. I will go ahead and merge, with the remark that we should probably add a delay to the update of the textures in the 3D ortho slices. That should happen before the next release.

PR #203, opened 2019-02-24
Cache recent directories

hanslovsky:
Opening datasets in Paintera can be annoying, even though there are a few conveniences such as remembering the last dataset that was opened during a session. However, that information is lost in subsequent or parallel sessions. This PR adds **experimental** features to make opening datasets more convenient:
 - load config from `~/.config/paintera.yml`:
   - default location for openers to look for data
   - favorite locations for openers
   - ... more in the future
 - cache recently opened containers in `~/.cache/paintera/<class>/recent`. Currently happens for n5 and hdf5 loaders (google loader not planned)
 - [fuzzy string matching](https://github.com/xdrop/fuzzywuzzy) node
 - Replace `Browse` button in n5 and hdf5 loads with `Open` `MenuButton with the following items:
  - `Browse`: same as the previous `Browse` button
  - `Recent`: read list of cached recent files and show to user for selection (with fuzzy matching)
  - `Favorites` read list of favorite locations from config file and show to user for selection (with fuzzy matching)

## Notes
  - These features are experimental and will probably remain so for a while after merging into master. 
  - Fuzzy matching could probably be improved. Current node is a workaround for non-dynamic JavaFX menus:
     - minimze the number of choices when typing:
        - `cutoff` - list only items with better score
        - `limit`    - list only that many items
     - https://stackoverflow.com/questions/54834206/javafx-dynamically-update-menu-while-showing
     - https://bugs.openjdk.java.net/browse/JDK-8219620
  - Fuzzy matching should also be added to dataset selection, especially useful for containers with many datasets, e.g. `lauritzen/02`.


hanslovsky:
Will merge soon.

It is pretty slow for containers with a looot of datasets (on NRS) but that's ok, I guess.

PR #209, opened 2019-03-09
Handlers per source only relevant changes for handlers per source

hanslovsky:
A global event handler that checks the type of the current source and then selects appropriate special handlers is very inflexible and cannot be extended for new source types (unless baked into Paintera core). Instead, add source-specific handlers into`SourceState` so any source can use custom handlers.


hanslovsky:
I toyed around with a lauritzen-02 project and I could still navigate, select, deselect, merge, split, refresh meshes, and the synapses would be rendered in 3D as well. I will go ahead and merge this PR.

PR #211, opened 2019-03-27
Support rendering of arbitrary intervals to speed up painting

igorpisarev:
This PR improves upon #200 and replaces tile rendering with arbitrary interval rendering. It significantly simplifies code, removes overhead, and makes painting feel more responsive. It also allows to handle painting over large areas better.

Changes:
* Completely remove render tile grid
* Keep track of requested repaint intervals at each screen scale level in `MultiResolutionRenderer`
* Modify volatile hierarchy projectors to accept `mask` as a RAI instead of an array and render only inside the relevant interval (TODO: submit a PR in bdv-core)
* Turn off adjusting `maxScreenScaleIndex` depending on how long the rendering takes in `MultiResolutionRenderer` as it does not work well with the new interval repaint logic.
* Use separate texture images for the orthoslices and copy partial updates of the render images into the texture images

## Demo

paintera-roi-based-renderer


hanslovsky:
Wow, this looks super awesome, thanks @igorpisarev! Unfortunately, I am super busy right now (and will be until June) and cannot go over the code right now. If you are confident you can merge without my review. I would like to try the PR first, though, and we should probably let it sit for a bit, anyway.

igorpisarev:
@hanslovsky It would be great if you could try it out! I've done a lot of testing while implementing it so I'm fairly sure that everything works properly, but a fresh look is always helpful. I'll probably go ahead and merge after that then.

igorpisarev:
As far as I know, @axtimwalde and Chris have been using this version for some time without any problems. Would it be a good time then to merge and make a new release? (I don't have write permissions in this repository)

hanslovsky:
Sounds reasonable to me, I will go ahead and merge, then.

> I don't have write permissions in this repository

That's my bad, sorry. I think I fixed it now. The easiest way to try is probably by creating a branch in the repository.

hanslovsky:
Once travis build succeeds on master, I will make a new minor version release (will probably not have time to release to conda before tomorrow).

igorpisarev:
Thanks @hanslovsky!

hanslovsky:
0.11.1 is now available on conda.

PR #218, opened 2019-04-23
Restrict width of the bottom status bar

### LINKED ISSUE #217: Restrict width of status bar (bottom)
The status bar prints, amongst other things, the voxel value under the cursor. For large multi-sets or many channels, this can be a very long string. The width of the status bar is (or seems to be) unconstrained, which leads to weird resizing and rendering artifacts. The fix should be easy: Restrict the width of the status bar (or its contents).

The easiest way to re-produce is with either a dataset with many channels (e.g. 12 channel affinity predictions) or downsampled label-multisets from high-variance label data and relatively small Paintera window size.

It is probably sufficient to restrict the width of the [status bar](https://github.com/saalfeldlab/paintera/blob/f5c43f51ff7360651bdf412eff06a79b05b87a88/src/main/java/org/janelia/saalfeldlab/paintera/BorderPaneWithStatusBars.java#L65), and possibly its children.

igorpisarev:
Fixes #217


hanslovsky:
Thanks @igorpisarev much appreciated!

PR #220, opened 2019-04-26
Mask fixes

igorpisarev:
Addresses the following problems:
* brush strokes not visible while painting
* painted brush strokes disappear when zooming in

## After

new

## Before 1

very-old

## Before 2

old

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,7 @@
 		<dependency>
 			<groupId>sc.fiji</groupId>
 			<artifactId>bigdataviewer-core</artifactId>
+			<version>6.0.2</version>
 		</dependency>
 		<dependency>
 			<groupId>com.google.code.gson</groupId>
--- a/src/main/java/bdv/fx/viewer/render/MultiResolutionRendererGeneric.java
+++ b/src/main/java/bdv/fx/viewer/render/MultiResolutionRendererGeneric.java
@@ -1054,10 +1054,22 @@ public class MultiResolutionRendererGeneric<T>
 		}
 	}
 
-	public synchronized void setScreenScales(double[] screenScales)
+	public synchronized void setScreenScales(final double[] screenScales)
 	{
 		this.screenScales = screenScales.clone();
 		createVariables();
+	}
+
+	/**
+	 * Set {@code screenScaleTransform} to a screen scale transform at a given {@code screenScaleIndex}.
+	 *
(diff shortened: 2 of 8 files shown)


hanslovsky:
Awesome, cool stuff, @igorpisarev 
Do you want me to check out the code?

igorpisarev:
@hanslovsky I am testing it now and going to push some more changes soon. After that it would be great if you could have a look! I'll let you know when it's ready

igorpisarev:
@hanslovsky I think this PR is now good to go.
I pushed one more change to fix discontinuities in rendered brush strokes. For that I changed the behavior of `getInterpolatedSource()` and `getInterpolatedDataSource()` in `MaskedSource`. Now it directly returns a real interpolated mask instead of interpolating a rastered source which resulted in subsampling values when other transforms were applied by the renderer.

To achieve that, I had to add quite a lot of classes duplicating existing functionality to make it work with real accessibles. (because `RandomAccessible` doesn't extend `RealRandomAccessible` in contrast to `Interval` and `RealInterval`).

Some of the added classes for real out-of-bounds handling can probably be moved to imglib2 later on.

### Before last commit:

before-last-fix

### After last commit:

after-last-fix

hanslovsky:
Thanks @igorpisarev, looks good from what I can see in this PR (haven't looked at the code/tried out yet). What is the performance penalty for the real out-of-bounds handling compared to integer, if any? I know that for integer values, the oob check is quite efficient. Is a similar optimization possible for the real coordinate case?

igorpisarev:
I copied and appropriately modified the integer-based out-of-bounds code , so it follows the same logic and should be pretty efficient too. I didn't benchmark it directly but at least didn't notice any performance difference when browsing data.

There is one more thing I'm thinking of: one possible way to reduce code duplication would be to only keep the real-based source composition classes and modify `getSource()` and `getDataSource()` to re-use their real counterparts and then do `Views.raster()` on top of them. But I thought that it might come at performance penalty because then each call to integer-based sources would have to go through the interpolator first. Maybe it's fine though, I'll test it and if it works well I'll probably push it too.

igorpisarev:
Generating integer-based sources via rastering real-based sources seems to work fine: tested navigating and painting and didn't noticed any problems nor difference in speed. And it really helps to avoid a lot of duplicated code that I added in the previous commit.
From my side it's OK to review/merge now.

PR #221, opened 2019-05-08
Selected label and mask status

igorpisarev:
* Display last selected label color in the status bar. Its ID is displayed in a tooltip on mouse over.
* Show progress indicator when applying mask to canvas

paintera-mask-status


igorpisarev:
Any source state can now display its status on the left side of the status bar by implementing `getDisplayStatus()` to return a custom JavaFX `Node`.

hanslovsky:
I just tried on my machine and it looks good to me, thanks a lot @igorpisarev 

Just two things:
 1 Please add a tool tip to the loading indicator.
 2 Update the (3D) flood fill (shift-F left click) to also use the loading indicator instead of changing the cursor. Click on the loading indicator should cancel the flood-fill (or open context menu to do so), if it is not applied into the canvas yet. This can be done in a separate PR, once this is merged.

Addendum to (2): The source could have an (observable) `Set<ObservableBooleanValue> busyStates` that indicate that something is busy. Then, the source itself would have a `BooleanBinding sourceIsBusy` that is `true` if `busyStates.tream().anyMatch(ObservableBooleanValue::get)`. `sourceIsBusy` would have to be updated whenever `busyStates` is modified.

Somewhat related: We could even go further and also have the source (state) provide the value under the cursor (currently that is handled globally). But that should happen in a separate PR.

igorpisarev:
Thanks for the ideas @hanslovsky!
I added a tooltip to the progress indicator showing the label ID of the mask. I agree that flood fill status would be great to have there too and will work on it in a separate PR.

hanslovsky:
Thanks a lot @igorpisarev

Issue #214, opened 2019-04-05
Cannot change interpolation to NLinear for raw data sources


igorpisarev:
Interpolation is always set to `NearestNeighbor` even for raw data sources. It's supposed to switch to `NLinear` on pressing `I`, but the interpolation property in `SourceState` remains the same so the change doesn't occur.

hanslovsky:
I wonder when that bug was introduced, maybe when restructuring the handlers.

hanslovsky:
This seems to work now, can I close this @igorpisarev?

igorpisarev:
Yes, it's been fixed there, closing now

PR #222, opened 2019-05-09
Toggle interpolation for all sources

igorpisarev:
Addresses #214
When the current source is `MaskedSource` that doesn't support linear interpolation, it can be confusing that nothing happens if you press `I`. I propose to toggle interpolation globally for all sources regardless of what the current source is.


src/main/java/org/janelia/saalfeldlab/paintera/PainteraDefaultHandlers.java
```diff
@@ -465,10 +465,11 @@ public static void grabFocusOnMouseOver(final Node node)
 
 	public void toggleInterpolation()
 	{
-		final Source<?> source = sourceInfo.currentSourceProperty().get();
-		if (source == null) { return; }
-		final ObjectProperty<Interpolation> ip = sourceInfo.getState(source).interpolationProperty();
-		ip.set(ip.get().equals(Interpolation.NLINEAR) ? Interpolation.NEARESTNEIGHBOR : Interpolation.NLINEAR);
+		for (final Source<?> source : sourceInfo.trackSources())
```
hanslovsky: This only makes sense if there is a global interpolation state. Think of this scenario:
Add source1, toggle interpolation (source1 is now in `NLINEAR`).
Add source2, toggle interpolation (source1 is now in `NEARESTNEIGHBOR`, source2 is now in `NLIENAR`).

igorpisarev: Good point, I'll look into changing interpolation mode to be a global state instead.

hanslovsky: We should keep the source specific interpolation (but control it through the global state for now), in case we want to be able to add an option to set interpolation individually per source

igorpisarev:
I have revisited this PR and have a proper fix for it, but it's implemented on top of #228, so I'll update this PR until that one is merged into master to avoid cluttering it with duplicating changes.

igorpisarev:
I've updated the PR. There is now a global interpolation state that ensures that all sources are in the same interpolation mode.

src/main/java/org/janelia/saalfeldlab/paintera/PainteraDefaultHandlers.java
```diff
@@ -441,11 +447,32 @@ public static void grabFocusOnMouseOver(final Node node)
...
 	public void toggleInterpolation()
 	{
-		final Source<?> source = sourceInfo.currentSourceProperty().get();
-		if (source == null) { return; }
-		final ObjectProperty<Interpolation> ip = sourceInfo.getState(source).interpolationProperty();
-		ip.set(ip.get().equals(Interpolation.NLINEAR) ? Interpolation.NEARESTNEIGHBOR : Interpolation.NLINEAR);
-		baseView.orthogonalViews().requestRepaint();
+		if (globalInterpolationProperty.get() != null)
+		{
+			globalInterpolationProperty.set(globalInterpolationProperty.get().equals(Interpolation.NLINEAR) ? Interpolation.NEARESTNEIGHBOR : Interpolation.NLINEAR);
```
hanslovsky: At some point we should use [`Interpolation.next`](https://github.com/bigdataviewer/bigdataviewer-core/commit/c9e036b3441047653693ddb176124ad0e75b3cf6) instead of cycling through interpolations manually. That requires update of BDV dependency version, which might break some code, so this should happen separately.

hanslovsky:
Looks good to me, will merge. In the future, we might have a toggle switch in each state to optionally interpolate independently of global interpolation state, but for now, this will do.

src/main/java/org/janelia/saalfeldlab/paintera/PainteraDefaultHandlers.java
```diff
@@ -441,11 +447,32 @@ public static void grabFocusOnMouseOver(final Node node)
...
+			baseView.orthogonalViews().requestRepaint();
+		}
+	}
+
+	private InvalidationListener createSourcesInterpolationListener()
+	{
+		return obs ->
+		{
+			if (globalInterpolationProperty.get() == null && !sourceInfo.trackSources().isEmpty())
```
hanslovsky: This would fail if `sourceInfo.trackSources().iterator().next()` has `null` for interpolation. I do not think that any sources with that behavior but we should be aware of this potential source of error.

hanslovsky:
Checked out the PR and confirmed that it works.

Thanks a lot @igorpisarev

PR #224, opened 2019-05-18
Channel selection in opener dialog

### LINKED ISSUE #172: Data open dialogue: set dataset type to label for label dataset
It would be nice if the type for a dataset that has the attributes `{"painteraData": {"type": "label"}}`
is automatically set to `label` instead of `raw` in the data open dialogue.

hanslovsky:
Currently, channel sources can only be added with all channels through the opener dialog; `PainteraShowContainer` can select individual channels. I added channel selection to the opener dialog to make channel data loading useful for scenarios such as affinity prediction w/long range affinities.

Fixes #172


hanslovsky:
Note that this also auto-detects the type of the selecte dataset (ignores the "painteraData" attribute, just checks for label multiset type/uint64/uint32/int64 for label, else raw -- could be improved)

hanslovsky:
Closes #172

PR #225, opened 2019-05-22
Allow to toggle ids with Ctrl+left click

igorpisarev:
I sometimes find myself trying to toggle label ids with `Ctrl`+left click similar to how one would select multiple files, for example, in a file browser application. This PR adds this shortcut in addition to the right click for toggling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ Usage: Paintera [-h] [--default-to-temp-directory] [--print-error-codes]
 | `Shift` + `M` | Maximize split view of one slicing viewer and 3D scene |
 | `Shift` + `Z` | Un-rotate but keep scale and translation |
 | left click | toggle id under cursor if current source is label source (de-select all others) |
-| right click | toggle id under cursor if current source is label source (append to current selection) |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -152,7 +152,7 @@ Usage: Paintera [-h] [--default-to-temp-directory] [--print-error-codes]
 | `Shift` + `M` | Maximize split view of one slicing viewer and 3D scene |
 | `Shift` + `Z` | Un-rotate but keep scale and translation |
 | left click | toggle id under cursor if current source is label source (de-select all others) |
-| right click | toggle id under cursor if current source is label source (append to current selection) |
+| right click / `Ctrl` left click | toggle id under cursor if current source is label source (append to current selection) |
```
hanslovsky: Typo here, replace `/` with `|`.

hanslovsky:
My only worry is that shift and ctrl are so close that users might accidentally merge. Shouldn't be a problem though because we can undo merge/detach actions. I will go ahead and merge.

Maybe we could have an undo shortcut as well (not in this PR), specific to each source. That would require some more consideration though (source-local undo? global undo? which actions should be undo-able? e.g. navigation?)

PR #228, opened 2019-06-03
Interpolation painting

igorpisarev:
Adds new shape interpolation mode that helps to automate painting over large areas.

There is a new block in the readme that explains the usage. The main key is `S` and is used for entering/advancing the mode. The mode is activated in the viewer window that has focus when the `S` key is pressed, and the rest of the controls and most of the shortcuts become inactive while in the mode. The rotation is disabled as well because interpolation between the sections requires them to be parallel.

Everything works as of now, and people can already start testing and using the new functionality.

There are a few ideas for further improvement, but they are not critical and can be addressed separately:
- [x] when previewing the results, allow to go back to the first/second section and change the selection
- [x] display the current status of the mode in the status bar (select first/second section, preview)
- [x] optimize how bounding box is determined for interpolation: currently it may be larger than necessary

------------------

### Demo

shape-interpolation


igorpisarev:
All the known/found bugs and edge cases are fixed now.
The hotkeys have also slightly changed: instead of having to press `S` several times, it's used only once to enter the mode. Then, the sections are switched automatically as soon as the user scrolls away from the current section that has a non-empty object selection. Committing the interpolated shape into the canvas is now activated by pressing `Enter`.

igorpisarev:
The final touches have been made and some more related issues have been fixed.
@hanslovsky Would it be OK with you to merge this PR into master? Everything has been tested quite thoroughly and works well.

Summary of changes:
* new shape interpolation mode, activated by pressing the `S` key
* the `Shift`+`M` shortcut now maximizes the currently active viewer window
* show both fragment and segment ids in the tooltip of the color indicator for selected label
* 2d/3d floodfill now considers the fragment-segment assignment and fills within the segment
* 2d floodfill bugfix: with increased thickness it used to fill outside the clicked object in some cases
* prevent application from freezing when still applying mask and a new paint action is requested
* sidebar fixes:
   * opened panel was stealing right arrow key presses preventing it from rotating the view
   * made the sidebar slightly wider and optimized layout of mesh settings to fit all elements

hanslovsky:
Sounds good to me, I have only one question: how is maximizing the viewer with `Shift` + `M` different from current behavior with just `M`?

Apologies for brevity, sent from phone.

igorpisarev:
`Shift`+`M` is still a 2D+3D side-by-side view, while `M` is still a fullscreen 2D view.
Sorry, I should've been clearer about this, the difference in the `Shift`+`M` shortcut is that now it puts the currently active viewer window side-by-side with the 3D viewer. Previously it was always using the current transform of the XY view for the 2D+3D view.

hanslovsky:
Gotcha. I think you had explained that to me in person at some point. I'll go ahead and merge.

igorpisarev:
Thanks @hanslovsky!

hanslovsky:
Thanks a ton for your great work @igorpisarev!! I haven't gotten around to try it myself but I will definitely advertise this new feature at the learnathon.

PR #235, opened 2019-06-20
Repaint crosshair overlays when layout changes

igorpisarev:
Before this fix the crosshair overlays wouldn't get repainted when the layout changes, causing them to look inconsistent with the new layout.


hanslovsky:
Great thanks a lot @igorpisarev

PR #236, opened 2019-06-20
Remove orthoslice delay setting

### LINKED ISSUE #227: Remove ortho-view delay setting from preference pane
This does not have any effect anymore as far as I can tell.

igorpisarev:
closes #227


hanslovsky:
Thanks a lot @igorpisarev

Issue #137, opened 2018-09-20
Add 'merge all selected' action


axtimwalde:
Add action that merges all selected segments.  Use the segment id of the last selected segment or the segment id of one of the selected segments if the selected segment is a fragment and has no segment id.  If all segments were fragments, invent a new segment id.

Optionally, execute as a sequence of pairwise merges with the last merge implementing the above specified segment id assignment.

PR #237, opened 2019-06-20
Add merge all selected action (Ctrl+Enter)

igorpisarev:
Adds a new shortcut `Ctrl`+`Shift`+`M` that will merge all selected fragments as described in #137.
(let me know if there are any better ideas for the key combination)

* If the last selected fragment belongs to a segment, its ID will be used for merging
* Otherwise, if the last selection doesn't belong to any assignment or is empty, the segment ID is chosen arbitrarily based on the assignment of the selected fragments (the order is not specified)
* If all selected fragments are trivial and are not part of any assignment, a new segment ID is created

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ Usage: Paintera [-h] [--default-to-temp-directory] [--print-error-codes]
 | `C` | Increment ARGB stream seed by one |
 | `Shift` + `C` | Decrement ARGB stream seed by one |
 | `Ctrl` + `Shift` + `C` | Show ARGB stream seed spinner |
-| `Ctrl` + `Shift` + `M` | Merge all selected ids |
+| `Ctrl` + `Enter` | Merge all selected ids |
 | `V` | Toggle visibility of current source |
 | `Shift` + `V` | Toggle visibility of not-selected ids in current source (if label source) |
 | `R` | Clear mesh caches and refresh meshes (if current source is label source) |
--- a/README.md
+++ b/README.md
@@ -170,6 +170,7 @@ Usage: Paintera [-h] [--default-to-temp-directory] [--print-error-codes]
 | `C` | Increment ARGB stream seed by one |
 | `Shift` + `C` | Decrement ARGB stream seed by one |
 | `Ctrl` + `Shift` + `C` | Show ARGB stream seed spinner |
+| `Ctrl` + `Shift` + `M` | Merge all selected ids |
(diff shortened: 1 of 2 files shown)


hanslovsky:
Thanks @igorpisarev
My only concern is that the shortcut is dangerously close to `M` and `Shift` `M` and I am worried about inadvertent merge all actions. Solutions:
 - new shortcut for merge all
 - new shortcuts for maximizing 2d viewers, eg F11 in analogy to full screen browser
 - no shortcuts for maximizing viewer, use gui elements instead, e.g. buttons in top right of each viewer
 - ignore those concerns and merge anyway

I think I would prefer options 2 or 3.

igorpisarev:
Using F11 for full screen is a good idea, but unfortunately I can't get it to work on a MacBook (F11 and fn+F11 are reserved in the OS). Web browsers on Mac have a different 3-key shortcut for full screen, which would overlap with the flood-filling shortcuts in Paintera, so using native Mac shortcuts is also not an option without changing something else.

Maybe let's keep `M` and `Shift+M`? Since they have been around for quite a while, users are probably already used to them. I really like the minimize/maximize buttons that you suggested in addition to the existing shortcuts.

As an alternative to protecting from accidental merge-all, we can add a confirmation dialog. It shouldn't be too annoying as long as the action is not used very frequently. The dialog can also list the details about the action (how many fragments are going to be merged, sizes of their assignments) and let the user pick the resulting segment ID.

hanslovsky:
A dialog sounds good to me, we could also have a checkbox to always apply without a merge, additionally.

igorpisarev:
Discussed it with @axtimwalde and decided to use a different shortcut for the merge-all action to keep things simple for now, so I changed it to `Ctrl`+`Enter` which doesn't overlap with anything else.
@hanslovsky What do you think?

The above suggestions would be nice to have, and we can add them separately later on. Also, the resulting segment ID determination for the merge-all operation can be improved: we can order the selected IDs in the chronological order and pick the most recently selected segment ID. (currently it is chosen arbitrarily if the last selected ID doesn't belong to any segment).

hanslovsky:
Thanks @igorpisarev!
Looks good to me now, ready to merge from your side?

igorpisarev:
Yes, I'll go ahead and merge then. Thanks!

hanslovsky:
Thanks @igorpisarev

PR #240, opened 2019-07-06
add icon and title

axtimwalde:
for gnome top bar title, more work will be necessary
(https://stackoverflow.com/questions/31219169/javafx-application-name-on-gnome)


hanslovsky:
Great draft @axtimwalde, looks really cool already. I have a few suggestions based on a screenshot when running on my laptop:
Screenshot_20190706_070751

 - Contrast of the blue eye not very high: A lighter/less saturated color (or white) would probably be good here.
 - Detail on the brush is lost: Only one color a the tip would be good, keep the brush gray scale other than that.
 - Consider removing white background circle

axtimwalde:
Good points.  I thought about refining over time as we learn how it pans out on various screens and platforms.  That's just a draft but better than the default icon on Ubuntu Gnome (which is an old computer screen with a no-parking symbol or whatever that is supposed to mean.  We will need some sort of background because it looks weird on dark background otherwise.  I have tried just outlining it with white but doesn't look good.  May be we outline the white circle with a black line, that could save it.   For the eyes and overall contrast, I thought some pseudo-3D effects would be nice but haven't found any that I like yet.

hanslovsky:
Updated and it looks better on my system:
Screenshot_20190706_172132

Does it still look good on Ubuntu? If so, I am ok with merge.

 - black outline on circle
 - lighter blue color for right eye
 - only one color on brush tip (blue
 - blue and magenta for brush handle and tip, respectively, because they go together nicely

axtimwalde:
Ok.  I took most of your changes and added pseudo-3D. Looks good on Ubuntu Gnome
image

hanslovsky:
Looks good on my Plasma with Breeze theme as well, will go ahead and merge.

PR #255, opened 2019-07-19
Bookmarks

### LINKED ISSUE #140: Add bookmarks
Like in BDV, enable storing and navigating to bookmarked viewer transforms.

hanslovsky:
Fixes #140
 - [x] add bookmark config
 - [x] add bookmark UIs
 - [x] add shortcuts to README
 - [x] serialize bookmarks into project file
 - ~[ ] add option to cancel applying bookmark and going back to previous state (within some time window)~
 - [x] add shortcut to go to bookmark (instead of going through UI)

Use these shortcuts to control bookmarks:
 -  `B`  Add bookmark with current global and 3D viewer transforms
 - `Shift` + `B` Open dialog to add bookmark with text note
 - `Ctrl`+`B` Open dialog to select bookmark

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,9 @@ Usage: Paintera [-h] [--default-to-temp-directory] [--print-error-codes]
 | `Ctrl` + `S` | Save current project state |
 | `Ctrl` + `Shift` + `N` | Create new label dataset |
 | `Ctrl` + `T` | Threshold raw source (only available if current source is raw source) |
+| `B`  | Add bookmark with current global and 3D viewer transforms |
+| `Shift` + `B` | Open dialog to add bookmark with text note |
+| `Ctrl`+`B` | Open dialog to select bookmark |
 
 ### Shape interpolation mode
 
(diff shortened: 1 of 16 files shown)


hanslovsky:
Undoing bookmark selection within a short time window after selection should be worked on in a separate PR if it ever becomes an issue.

igorpisarev:
Super-useful feature @hanslovsky!
Can you also add the bookmark actions as enum entries in one of the action types here (or as a new action type)?
https://github.com/saalfeldlab/paintera/tree/master/src/main/java/org/janelia/saalfeldlab/paintera/control/actions
And then check if the action is allowed in the predicate for the bookmark key shortcuts like this: https://github.com/saalfeldlab/paintera/blob/28cda971ee4c950b92472a28031eb3fdd4541b50/src/main/java/org/janelia/saalfeldlab/paintera/PainteraDefaultHandlers.java#L379
This would disable the bookmark actions while in the shape interpolation mode, where it's important that the viewer transform doesn't change arbitrarily.

hanslovsky:
@igorpisarev I added a new `NavigationActionType.Bookmark` `enum` entry and it seems to work as expected.

hanslovsky:
I will go ahead and merge.

Issue #254, opened 2019-07-19
(Arbitrary) Mesh Import


hanslovsky:
@cpatrickc made an awesome video using catmaid demonstrating the size of the cremi and lauritzen blocks compared to the entire fafb volume. Starting from that we were brainstorming how we could use this visualization to then zoom into one of the lauritzen blocks and show all the cool work that @cpatrickc is doing there.

For that we would need to be able to load arbitrary meshes into Paintera. Note that this is entirely independent of currently open sources and should be managed through the general settings inside the preference pane, and serializable. Each mesh should have settings color, cull face, etc.

Does this intersect with your work on visualizing skeletons @axtimwalde? Or are you generating the skeletons within Java?

Based on this idea, @cpatrickc also said that there are polygons/triangle meshes that delineate potential areas of interest. It would be super cool to import these (like any mesh above) and then have an overlay on the screen that indicates which portions of the screen are inside an area of interest.

 - [ ] arbitrary mesh import
 - [ ] overlay to indicate which parts of screen are inside imported polygon/triangle mesh

PR #262, opened 2019-07-24
Mesh openers

hanslovsky:
I added support for loading arbitrary meshes into the Paintera viewer. @cpatrickc this will allow you to import the fafb landmark mesh into Paintera.

Example image with examples `alfa147.obj` and `al.obj` from https://people.sc.fsu.edu/~jburkardt/data/obj/obj.html:

Screenshot_20190724_174628

The UI in the side bar has a new entry `Triangle Meshes` and new meshes can be added with the `+` button.

Screenshot_20190724_174918

Addresses the first item of #254

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -157,6 +157,13 @@
 			<version>2.0</version>
 		</dependency>
 
+<!--		Load meshes-->
+<!--		https://github.com/javagl/Obj -->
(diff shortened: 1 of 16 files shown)


axtimwalde:
Material should become the phong material used for label meshes though.

PR #267, opened 2019-07-25
Adaptive mesh resolution

igorpisarev:
Adds support for adaptive level of detail for 3D objects, where each block is rendered at appropriate resolution. The scene is updated whenever the orientation of the camera changes (throttled to only one update every 3 seconds).

The appropriate level of detail is picked automatically, and the new settings allow to tweak it:
* **Preferred scale level** sets a threshold on how many screen pixels a projected voxel can occupy. The number of pixels for the given scale level is determined by the mipmap factors (usually it's `2^level`). If the projected voxel occupies more pixels than that, the block is subdivided to increase the level of detail.
* **Highest scale level** simply sets the limit on what the highest resolution can be.

The size of the blocks is specified by the new 3D viewer setting **renderer block size**. This overrides the source block sizes and essentially sets the length of the 3D block along X/Y, and the actual block size in Z is determined automatically for each source based on the resolution ratio. This helps to make the block sizes close to isotropic at all scale levels.

## Summary
* View frustum culling to filter out blocks that are outside the screen
* Seamless scene updates: re-rendered blocks are replaced on-the-fly while browsing
* New mesh settings: preferred scale level and highest scale level
* New 3D viewer settings: renderer block size and whether to display outline of rendered blocks (adds FXyz library for 3D boxes made up of quadrilateral faces)

## Demo
paintera-multires-meshes-cremi-3


igorpisarev:
@axtimwalde @hanslovsky @cpatrickc
I would greatly appreciate it if you could test it on your data! I've been testing it on a couple of my data sets, but it would be interesting to see how it performs on something else.

axtimwalde:
It seems that the queue of updates is not updated but appended?  If I zoom out and quickly in again, all meshes are first replaced by low resolution meshes and then again by high resolution meshes.  Instead, when zoomed in the updates to low resolution meshes should not be done.  Observing this with a large selection of about 100 meshes.

hanslovsky:
I will try on my machine @igorpisarev

igorpisarev:
This seems to be because of submitting too many tasks to the FX application thread (JavaFX requires to modify the scene on the main thread). Meshes are generated in the background and then submitted to the application thread to actually be added onto the scene, and these updates may pile up and lead to unnecessary additions and removals. To fix it I added a check if the mesh is still relevant before uploading it onto the scene.

I'm also experimenting with throttling the mesh updates on the FX thread to allow UI processing in-between those updates. Currently the application can become unresponsive if you're in the middle of a very dense structure. I'll update the branch when I have a working solution for this.

axtimwalde:
Great! I guess, meshes that have been generated should be saved in mesh cache, but not necessarily put into the scene.

On July 27, 2019 1:03:30 AM GMT+02:00, Igor Pisarev <notifications@github.com> wrote:

This seems to be because of submitting too many tasks to the FX application thread (JavaFX requires to modify the scene on the main thread). Meshes are generated in the background and then submitted to the application thread to actually be added onto the scene, and these updates may pile up and lead to unnecessary additions and removals. To fix it I added a check if the mesh is still relevant before uploading it onto the scene.

I'm also experimenting with throttling the mesh updates on the FX thread to allow UI processing in-between those updates. Currently the application can become unresponsive if you're in the middle of a very dense structure. I'll update the branch when I have a working solution for this.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_saalfeldlab_paintera_pull_267-3Femail-5Fsource-3Dnotifications-26email-5Ftoken-3DAAC6AMBSL37JA4WTIQGNXA3QBN7EFA5CNFSM4IG6YMN2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD254Y3Q-23issuecomment-2D515624046&d=DwMCaQ&c=LU6cRtx0xgB8s29tIz9Olw&r=kg2ubjZkQqptXEUbX_Qfk_IFQhpbRWyBPgrInYl05ko&m=TSaKeXDD1DRMcdCmtSkcx_vSO9UcOgnRjD4fdFjpKrA&s=hWqtu4DvC2Of1gk122oBTzbNNIlGrx0VcAQ1wp_9Zck&e=>, or mute the thread<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_notifications_unsubscribe-2Dauth_AAC6AMHPCCHWLJAKQY2DI7LQBN7EFANCNFSM4IG6YMNQ&d=DwMCaQ&c=LU6cRtx0xgB8s29tIz9Olw&r=kg2ubjZkQqptXEUbX_Qfk_IFQhpbRWyBPgrInYl05ko&m=TSaKeXDD1DRMcdCmtSkcx_vSO9UcOgnRjD4fdFjpKrA&s=Pe3XuQP28PZkYKUr47_1xvURUKnjV8s3tlNtPut2QUY&e=>.

--
Sent from my Android device with K-9 Mail. Please excuse my brevity.

igorpisarev:
Added:
* Prioritized mesh generation: blocks that are closer to the camera are generated first.
* Slightly different logic for scene updates: infrequent updates while navigating (every 5s) and forced update right after navigating (0.5s). This reduces slowdowns while browsing the scene and also updates the scene sooner once the camera is fixed.

This together should make 3D browsing a bit more responsive. However, my idea with throttling updates on the FX thread didn't work: adding/removing children is very fast and is not the bottleneck here. The occasional slowdowns seem to be caused by actual re-rendering of the scene when a lot of stuff has been added, but it happens in JavaFX's internals and we don't have any control over it.

@axtimwalde @hanslovsky The PR is now ready to be considered for merging. I don't think there is a lot of room for improvement left. The only thing that seems to be not very intuitive is the preferred scale level setting, which specifies how many screen pixels any given source voxel should occupy. Let me know if you have better ideas how to call/present this setting, otherwise I think it's good to go.

axtimwalde:
This PR should not be merged at this time.  The goal of this branch is to accelerate 3D rendering for large objects or selections.  In a test project with many objects selected, the 3D rendering speed of this branch is at all times significantly slower than without adaptive meshes.  While the selection is large, my laptop can fit all meshes onto the GPU.  Once there, 3D rendering is smooth and fast, individual meshes drop in slowly, but 3D rendering speed is ok while this happens.  With this branch, every motion in the 3D viewer triggers updates and these updates make the 3D viewer act like stop motion.  I.e. this feature needs further attention before it can be accepted.  From the user perspective, at this time, no adaptive mesh resolution is always better than adaptive mesh resolution.

axtimwalde:
@pisarev @hanslovsky, it seems that the mesh list in the side panel is a significant contributor to slowness.  Each mesh currently has a drop down thing and a progress bar, a button to export it individually and individual settings if necessary.  It seems that this large set of UI controls makes Paintera very slow.  It would be great if you could try how it works without the individual mesh controls and UI elements and the related property listeners.  If that makes it much faster, we may consider to maintain only a global progress bar for all meshes (helps me more than a seemingly infinite list of individual progress bars that flicker occasionally), and we could have UI for individual properties (if desired) that is created for only one mesh at a time as selected, e.g. in the 3D viewer. It may juts be that JavaFX does not handle hundreds or thousands of complicated UI elements with their property listeners well.  Having them all up is also not very useful to do anything with it.

hanslovsky:
I am OK with removing the mesh status entirely (or keep a central one that tracks global progress). We might be able to remove the `controlsFX` dependency then, as well. Creating a UI for individual settings only if they are actually used seems reasonable, too.

@axtimwalde How did you test the speed without those UI elements? Did you remove the side panel from the UI and ran Paintera that way?

igorpisarev:
@axtimwalde I tested it without the individual mesh controls on a selection of ~100 objects (CREMI), and I didn't notice any major changes in performance. For that I just commented out the listener in `MeshInfos` class, so the individual controls tab is always empty.

All scene updates are executed on background threads and don't require synchronization with the application thread. The main reason for slowness is uploading the generated meshes onto the GPU, which is done by JavaFX behind the scenes. It's triggered by adding the new objects to the scene's children on the application thread, but this operation is very fast and the actual work is done afterwards.
It doesn't seem like there is a good solution for this other than decreasing the number of 3D objects by tweaking the preferred & highest scale level settings.

axtimwalde:
@hanslovsky I did not test without the UI elements.

axtimwalde:
@pisarev do you have issues with performance at all?  May be you're testing in a regime where everything is fine, with or without UI elements.  Also, I was hoping for a test without the UI elements, i.e. no MeshInfo.  Just making inactive may still run code that makes things slow, not sure how JavaFX updates the UI when there are many many nodes.

igorpisarev:
@axtimwalde I'm experiencing occasional slowness if I navigate deep into the structure and then let it generate the meshes (I've set the highest scale level to 1, so the blocks that are closest to the camera are generated at this resolution). It gets stucky for 1-2 seconds when it needs to replace the old low-res meshes with the new high-res meshes. The dataset I'm testing it on (membranes) contains only about 10 labels, but each label occupies almost the entire volume.

So the `MeshInfo` UI doesn't seem to be an issue here (although turns out it was in [select-all](https://github.com/saalfeldlab/paintera/tree/select-all)!)
My understanding is that the slowness here is due to moving too much data onto the GPU at once and the way how JavaFX re-renders the scene after adding new nodes.

For example, here is a thread dump while the application was busy with replacing the meshes:
```
"JavaFX Application Thread" #27 prio=5 os_prio=0 tid=0x00007fe8ec336800 nid=0xd9fb waiting on condition [0x00007fd99fffd000]
   java.lang.Thread.State: WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00007fe52397f960> (a java.util.concurrent.CountDownLatch$Sync)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:997)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1304)
        at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:231)
        at com.sun.javafx.tk.quantum.PaintCollector.waitForRenderingToComplete(PaintCollector.java:157)
        at com.sun.javafx.tk.quantum.GlassScene.waitForRenderingToComplete(GlassScene.java:127)
        at javafx.scene.Scene$ScenePulseListener.pulse(Scene.java:2410)
        at com.sun.javafx.tk.Toolkit.lambda$runPulse$34(Toolkit.java:398)
        at com.sun.javafx.tk.Toolkit$$Lambda$1083/753866232.run(Unknown Source)
        at java.security.AccessController.doPrivileged(Native Method)
        at com.sun.javafx.tk.Toolkit.runPulse(Toolkit.java:397)
        at com.sun.javafx.tk.Toolkit.firePulse(Toolkit.java:424)
        at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:518)
        at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:498)
        at com.sun.javafx.tk.quantum.QuantumToolkit.pulseFromQueue(QuantumToolkit.java:491)
        at com.sun.javafx.tk.quantum.QuantumToolkit.lambda$runToolkit$291(QuantumToolkit.java:319)
        at com.sun.javafx.tk.quantum.QuantumToolkit$$Lambda$45/587463907.run(Unknown Source)
        at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
        at com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
        at com.sun.glass.ui.gtk.GtkApplication.lambda$null$203(GtkApplication.java:139)
        at com.sun.glass.ui.gtk.GtkApplication$$Lambda$41/2027961269.run(Unknown Source)
        at java.lang.Thread.run(Thread.java:748)

   Locked ownable synchronizers:
        - None
```
The application thread for some reason needs to wait until the newly uploaded objects are rendered. Maybe JavaFX was not designed to handle heavy on-the-fly updates effectively.

igorpisarev:
I've reworked the branch using a different approach for updating meshes in the scene.

The bottleneck turned out to be how JavaFX transfers the meshes onto the GPU, which led to blocking the FX thread. To avoid it,  I significantly decreased the renderer block size to make individual meshes much smaller (this is independent from the source block size and can be adjusted in 3D viewer settings). I also added a queue for the meshes that need to be added onto the scene that limits the total number of elements per frame (vertices+normals+faces). This helps to control the update rate and keep the FX thread responsive. Both of the queue parameters such as total number of elements per frame and delay between two frames can also be adjusted in 3D viewer settings.

There are a few more things that changed:
* Rendering is now hierarchical and will render blocks at intermediate scale levels first before replacing them with higher-resolution version. For example, when a user quickly zooms in, this approach allows to gradually improve resolution rather than having to wait until all highest-resolution blocks are rendered without seeing any updates.
*

igorpisarev:
I've reworked the branch using a different approach for updating meshes in the scene.

The bottleneck turned out to be how JavaFX transfers the meshes onto the GPU, which led to blocking the FX thread. To avoid it,  I significantly decreased the renderer block size to make individual meshes much smaller (this is independent from the source block size and can be adjusted in 3D viewer settings). I also added a queue for the meshes that need to be added onto the scene that limits the total number of elements per frame (vertices+normals+faces). This helps to control the update rate and keep the FX thread responsive. Both of the queue parameters such as total number of elements per frame and delay between two frames can also be adjusted in 3D viewer settings.

There are a few more things that changed:
* Rendering is now hierarchical and will render blocks at intermediate scale levels first before replacing them with higher-resolution version. For example, when a user quickly zooms in, this approach allows to gradually improve resolution rather than having to wait until all highest-resolution blocks are rendered without seeing any updates.
* The scene update is initiated after a short delay once the user stops navigating. This delay can also be adjusted in 3D viewer settings.
* Updating the block tree now works differently and is very fast, so if a user paints something, the block tree is updated immediately, and the 3D scene can be re-rendered without any lags.
* The blocks that are closer to the camera will be rendered first.
* There is now a global mesh progress displayed on the "Mesh List" tab.

More testing is needed, but in general the performance now is much better than before.

igorpisarev:
For some reason I sometimes get the following exception. It's a rather rare occurrence, but once it's happened the only way is to restart the app, otherwise it will just keep happening:
```
Exception in thread "JavaFX Application Thread" java.lang.ArrayIndexOutOfBoundsException: -1
	at java.util.ArrayList.elementData(ArrayList.java:422)
	at java.util.ArrayList.get(ArrayList.java:435)
	at com.sun.javafx.collections.ObservableListWrapper.get(ObservableListWrapper.java:89)
	at com.sun.javafx.collections.VetoableListDecorator.get(VetoableListDecorator.java:306)
	at javafx.scene.Parent.updateCachedBounds(Parent.java:1591)
	at javafx.scene.Parent.recomputeBounds(Parent.java:1535)
	at javafx.scene.Parent.impl_computeGeomBounds(Parent.java:1388)
	at javafx.scene.Node.updateGeomBounds(Node.java:3577)
	at javafx.scene.Node.getGeomBounds(Node.java:3557)
	at javafx.scene.Node.getLocalBounds(Node.java:3478)
	at javafx.scene.Node.updateTxBounds(Node.java:3641)
	at javafx.scene.Node.getTransformedBounds(Node.java:3424)
	at javafx.scene.Parent.getChildTransformedBounds(Parent.java:1732)
	at javafx.scene.Parent.updateCachedBounds(Parent.java:1596)
	at javafx.scene.Parent.recomputeBounds(Parent.java:1535)
	at javafx.scene.Parent.impl_computeGeomBounds(Parent.java:1388)
	at javafx.scene.Node.updateGeomBounds(Node.java:3577)
	at javafx.scene.Node.getGeomBounds(Node.java:3530)
	at javafx.scene.Node.getLocalBounds(Node.java:3478)
	at javafx.scene.Node.impl_intersectsBounds(Node.java:5013)
	at javafx.scene.Parent.impl_pickNodeLocal(Parent.java:705)
	at javafx.scene.Node.impl_pickNode(Node.java:4912)
	at javafx.scene.SubScene.pickRootSG(SubScene.java:806)
	at javafx.scene.SubScene.impl_pickNodeLocal(SubScene.java:824)
	at javafx.scene.Node.impl_pickNode(Node.java:4912)
	at javafx.scene.layout.Region.impl_pickNodeLocal(Region.java:2936)
	at javafx.scene.Node.impl_pickNode(Node.java:4912)
	at javafx.scene.layout.Region.impl_pickNodeLocal(Region.java:2936)
	at javafx.scene.Node.impl_pickNode(Node.java:4912)
	at javafx.scene.layout.Region.impl_pickNodeLocal(Region.java:2936)
	at javafx.scene.Node.impl_pickNode(Node.java:4912)
	at javafx.scene.Scene$MouseHandler.pickNode(Scene.java:3899)
	at javafx.scene.Scene$MouseHandler.access$1600(Scene.java:3485)
	at javafx.scene.Scene.pick(Scene.java:1942)
	at javafx.scene.Scene.access$6700(Scene.java:159)
	at javafx.scene.Scene$MouseHandler.process(Scene.java:3711)
	at javafx.scene.Scene$MouseHandler.access$1500(Scene.java:3485)
	at javafx.scene.Scene$MouseHandler$1.run(Scene.java:3521)
	at com.sun.javafx.application.PlatformImpl.lambda$null$172(PlatformImpl.java:295)
	at java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$173(PlatformImpl.java:294)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
```
This seems to occur in JavaFX internals, as there is no our code in the stack trace. There is a chance that it's initiated by our code, but it's unlikely because the scene can be updated only on the FX thread. I found a very similar bug report that is marked as unresolved: https://bugs.openjdk.java.net/browse/JDK-8163078
It appears to be a random issue, so I'll have to look into it more to find out when it happens exactly.

igorpisarev:
I've improved the branch since the last update with simplified code and proper handling for some edge cases that weren't covered before. All bugs also seem to be fixed now.

@axtimwalde Can you please give it a try once more on your data when you get a chance? I'll let this branch sit for a little bit before merging to make sure that it's fully tested.

igorpisarev:
I forgot to mention that I had to disable the mesh context menu for centering the orthoslices at the clicked location. This turned out to be another cause for the initial slowness: JavaFX internally recomputes `PickRay`/`PickResult` for some of the `MeshView` objects in the scene on every mouse move. This becomes rather slow when the number of objects is very high. I had to do `MeshView.setDisable(true)` for each object in the scene to switch off this behavior, which in turn disabled the 3D viewer context menu.

axtimwalde:
Nice, I am currently testing.  Isues so far:

1. Context menu to center ortho-slices is super important and must come back but must calculate ray intersection on mouse right click and not on mouse move.
2. Select all seems to trigger all objects to be generated at lowest resolution with highest priority even if the camera sees only three of them.  This means that I will wait forever (8000 objects) before I can look at the three objects in the camera at high resolution.  Solution: prioritize what's visible over everything else.  When camera moves, reprioritize.

axtimwalde:
3. It generally seems to load a lot of stuff that I am not currently looking at.  It also seems that this stuff has priority and that the computer is always busy, slowing down navigation speed while not updating anything relevant in the current view.

igorpisarev:
>     1. Context menu to center ortho-slices is super important and must come back but must calculate ray intersection on mouse right click and not on mouse move.

Good point, I'll look into it in more detail.

>     2. Select all seems to trigger all objects to be generated at lowest resolution with highest priority even if the camera sees only three of them.  This means that I will wait forever (8000 objects) before I can look at the three objects in the camera at high resolution.  Solution: prioritize what's visible over everything else.  When camera moves, reprioritize.

Some of the pending tasks are already reprioritized when the camera moves, but only those that have not been scheduled yet (i.e. higher resolution blocks because they depend on their low-res parent and will be scheduled once it's ready). When you select all, it indeed generates and submits all low-res tasks with their initial priority, so it doesn't change after the camera moves. I'll explore the options to reprioiritize submitted tasks, possibly stopping the task and starting a new one, or something more efficient such as maintaining the task queue ourselves.

>     3. It generally seems to load a lot of stuff that I am not currently looking at.  It also seems that this stuff has priority and that the computer is always busy, slowing down navigation speed while not updating anything relevant in the current view.

This is likely a consequence of the previous point, so this should become better when all tasks are reprioritized. But yes, it still does render blocks outside the screen but only at the lowest possible resolution without subdividing them further -- this is to be able to display the entire structure as a combination of blocks at different resolutions. Otherwise if we simply ignore the blocks outside the screen, when you zoom out such that the previously hidden part becomes visible, it will be empty until the requested meshes are generated.

axtimwalde:
May be jobs can make a priority check before they actually start (still high priority ? run : reschedule self).  Also, rendering non-visible blocks at the lowest resolution in a world with 8000 objects means that 8000 objects are being meshed while I am looking at only 3 of them.  I do not think this is a good idea.  Some padding of the visible space sounds reasonable but everything is too much.

src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
```diff
@@ -460,6 +457,8 @@ public void invalidateCaches()
 	{
 		Stream.of(this.blockListCacheInvalidate).forEach(InvalidateAll::invalidateAll);
 		Stream.of(this.meshCacheInvalidate).forEach(InvalidateAll::invalidateAll);
+		if (labelBlockLookup instanceof LabelBlockLookupFromN5)
```
hanslovsky: I'd rather add this to the `LabelBlockLookup` interface than doing an `instanceof` check for a specific implementation. May be useful to implement the entire `Invalidate` interface that imglib2-cache now provides.

src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
```diff
@@ -123,6 +118,7 @@
 
 	public MeshManagerWithAssignmentForSegments(
 			final DataSource<?, ?> source,
+			final LabelBlockLookup labelBlockLookup,
 			final Pair<? extends InterruptibleFunction<TLongHashSet, Interval[]>, Invalidate<TLongHashSet>>[] blockListCacheAndInvalidate,
```
hanslovsky: Passing the `labelBlockLookup` makes `blockListCacheAndInvalidate` obsolete, if I am not mistaken.

igorpisarev: For N5 backend yes, but currently `blockListCacheAndInvalidate` is a traditional `imglib2-cache` for `Label ID -> block lists` mapping, which is still helpful for example in case of file-based lookup where bulk requests cannot be performed more efficiently than individual queries. The new N5-specific cache allows to request a big bulk of block lists and store them which works behind the imglib2-based cache for individual entries.

I also find this logic a bit convoluted, but currently don't see a better way to put these two caches together if we want to be efficient and still support multiple label-to-block lookup formats.

src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
igorpisarev: Good point, I didn't know that `Invalidate` was an interface in imglib2-cache, it's probably in one of the newer versions. Although it's a bit more granular than needed here, I'll try to make use of it. Maybe we could add another interface that would extend `LabelBlockLookup` and `Invalidate` meaning that it's a lookup with added cache functionality, and let N5-based block lookup implement it, and keep the rest as is.

src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
hanslovsky: We could just add the imglib2-based cache into the individual implementations of `LabelBlockLookup` (or no cache at all if appropriate), no?

src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
hanslovsky: Yes, it is in a more recent version that the [remove-global-cache branch](https://github.com/saalfeldlab/paintera/pull/316) uses. Updating to that more recent version was quite some work so it might be good to way until that branch is merged.

igorpisarev:
I have made a lot of changes to the branch again mainly to increase performance and also add some missing features. The changes include:
* Re-prioritization of all pending tasks when the camera moves. This applies to both the render queue and the FX queue (the one that uploads the meshes onto the scene / GPU).
* The block tree is now global and is built for the entire scene when the camera moves instead of a separate block tree for each label. It's based on the global mesh settings so usually it's just one tree for the entire set of meshes, but if the user sets individual mesh settings that differ from the global settings, a separate block tree will be built for them.
* Speed up initial task creation when selecting thousands of objects by caching N5 label-to-block lookup. (https://github.com/saalfeldlab/label-utilities-n5/pull/3)
* Make UI more responsive when rendering thousands of objects by optimizing and throttling updates of the global progress bar on the FX thread.
* Added new mesh setting **Min label ratio** that specifies how the mask is generated from the label multiset data. If it's 0, it will only do the `contains()` test (old behavior). If it's higher, let's say 0.5, it will check that at least 50% of the contained labels in a pixel are the lables that we're interested in (that belong to the current segment). This helps to smooth the transitions between the parts of an object rendered at different resolution.
* Revised the mesh settings to **Level of detail** that sets how many pixels a projected voxel can have at max, and **Highest scale level** which is simply a limit on maximum possible resolution (same as before).
* Bug fixes in block tree intersection algorithm that previously resulted in stalled low-res blocks and missing tasks.

The current state of the branch works much better with large sets of objects. I think it can be considered for merging now, but it would be nice to test that there are no memory leaks like the one that was fixed in #340. I merged the master branch with the fix into this PR, but there has been a lot of changes and merge conflicts, so some extra validation is needed.

igorpisarev:
One more thing to note is that the mesh generation task queue may significantly grow over time when there is a lot of selected objects and the user moves the camera often. This is because of how the tasks are re-prioritized and the limitations of the priority-based task executor. So far I haven't observed any problems during testing, but this can make the rendering slower after prolonged use. But this is not a major issue and can be addressed separately.

More detailed explanation: our custom `PriorityExecutorService` uses standard `PriorityBlockingQueue` which doesn't allow to change priorities of existing elements. Removing an old element and inserting it again with a new priority is also not efficient because removing an element has linear complexity `O(n)`. So in order to re-prioritize efficiently the following approach is used:
* When the new priority is lower than the one that the task was scheduled with, the initially scheduled task will run and reschedule itself with lower priority.
* When the new priority is higher than it was scheduled with, the task is scheduled again so it can run sooner, and when the initially scheduled task runs at some point later in time, it checks if it's already been executed and exits without doing anything.

A better way to do this may be to use the custom worker thread pool and a custom priority queue that allows efficient updates, i.e. the one that is used for the FX queue: https://github.com/saalfeldlab/paintera/blob/0479d997806056b151804549063444b091f7d3d3/src/main/java/org/janelia/saalfeldlab/util/HashPriorityQueue.java
Doing it this way turned out to be not so trivial because the task synchronization logic would need to change a little bit. Also, there may be a performance trade-off because `PriorityBlockingQueue` can better deal with thread congestion.

igorpisarev:
I have implemented a new mechanism for queueing and running mesh generation tasks that allows to efficiently modify priorities of existing tasks. It feels a little bit more responsive now when doing `Ctrl+A` and turning the camera a lot while the meshes are being generated.

I also compared the task queue size in both implementations, and the old queue is growing over time which is expected, while the new queue is always bounded. The plot shows the queue sizes after 3 minutes of browsing with all objects being selected (these are two different runs on the same dataset with the same mesh settings and browsing patterns):

queue-size comparison

igorpisarev:
Unfortunately there seems to be another bug in the block tree intersection algorithm: occasionally the progress bar will stay incomplete even though there are no more active tasks. This is a separate issue from the missing tasks that I had earlier and occurs because of the disagreement between the requested block tree and the resulting block tree (which should be gradually transformed from the current state of the tree into the requested tree). I don't understand yet where this discrepancies come from, and it occurs intermittently and only on a dataset that has very large objects resulting in thousands of blocks, so it's very hard to debug.

As far as I can tell it doesn't have anything to do with the new task queue because it occurs on 
0479d99 before the changes.

igorpisarev:
I made a final round of fixes and improvements which include:
* Bug fixes in the block tree update algorithm and much stronger assertion logic to ensure correctness.
* New **coarsest scale level** setting that allows to change the lowest resolution of the blocks (also renamed the highest resolution setting to **finest scale level**). Together these two settings allow to replicate the old behavior where the meshes will be rendered at just the given scale level if set to the same value.
*Note that the rendering will still start from the lowest possible resolution, but then the low-res blocks will be gradually replaced with higher-res blocks until the specified coarsest scale level requirement is met.*
* The block tree for the scene is now built on a worker thread instead of the FX thread. Usually it's quite fast, but if the user sets the coarsest scale level to 0 (meaning that all blocks must be rendered at the highest possible resolution), the block tree construction may take some time. In this case it's helpful to not block the application thread and allow to interrupt the operation if the user changes the settings back to less extreme values.

igorpisarev:
I've made one more improvement: rendered blocks are now kept in the scene if they're outside the screen.

Previously, if the user turned the camera such that some of the rendered blocks went outside the screen, they were removed from the scene. Then if the user returned the camera to its previous position, these blocks would become visible again but they were already removed from the scene, so they needed to be fetched from the cache and added to the scene again. This doesn't happen immediately, so it looked like parts of the object disappeared for a short period of time.

Now, already displayed blocks will always stay in the scene, even if they're outside the screen. This helps to avoid the above problem and makes the navigation smoother. But, when pending blocks go outside the screen, mesh generation will be stopped for them so there's no unnecessary computation. The processing will be restarted for these blocks if they go inside the screen again later on.

igorpisarev:
I've compared the running times of generating ~6000 objects in a single dataset between this branch and the current master:
* On my workstation:
  * this branch: 64px blocks: **50s**,  32px blocks: **1m 10s**,  16px blocks: **1m 30s**
  * `master`: 64px blocks (same as source): **4m 30s**
* On my laptop (dual-core CPU with hyperthreading):
  * this branch: 64px blocks: **5m 10s**,  32px blocks: **5m 30s**,  16px blocks: **8m 10s**
  * `master`: 64px blocks (same as source): **5m 30s**

The major speedup on the workstation comes from the new caching layer in N5 label-to-block lookup (now the lookup file is read only once and cached, while in the master branch it's accessed for every label id and causes a lot of overhead due to file locking on a network file system).

On the laptop the data is stored locally on a SSD so there is not much overhead in accessing the same lookup file multiple times, and the runtimes are comparable with the same block size.
(actually in this branch the block sizes are adjusted in Z to make them closer to isotropic, so the actual block sizes are 64x64x53 for 64px,  32x32x27 for 32px,  and 16x16x13 for 16px)

For some reason the laptop run of this branch is significantly slower with the 16px block size. This may be due to the increased number of objects in the JavaFX scene, or due to the padding in the marching cubes algorithm for each rendered block which increases the size of the meshes. All tests were done using the same settings: the lowest resolution scale level and no mesh simplification/smoothing.

igorpisarev:
I've finished merging this branch with the most recent master and adapting to the changes introduced in #343. Both `LabelSourceState` and `ConnectomicsLabelState` support adaptive meshes now, and I've also verified that all functionality added in this branch still works as expected.

However, the merge process was quite difficult to get through. This branch needs to be merged into master before making any other significant changes to the code base (perhaps after the next release that will add relative path support).
This PR works very well on both my workstation and my laptop and improves 3D scene interactivity in many ways. The default block size for the renderer is 16px, and while it may be slightly slower than 32px or 64px for certain datasets (such as with relatively small objects like cells because of the overhead in having more objects in the scene), it's adjustable so the user can tweak the settings a little bit to see what works best for a particular dataset on their particular device. I've found that 16px gives good results on most datasets that I've tested it with while still allowing for high granularity of the 3D scene.

hanslovsky:
Thanks a ton @igorpisarev 
This is awesome!!

PR #268, opened 2019-07-26
Use MultiResolutionRenderer from bigdataviewer-core

maarzt:
The MultiResolutionRendererGeneric in Paintera is a fork of the MultiResolutionRenderer in BigDataViewer. This PR is part of an effort. That tries to reintegrate the changes made in Paintera into BigDataViewer.

See also https://github.com/bigdataviewer/bigdataviewer-core/pull/59

Currently I try to reduce the differences between the MultiResolutionRenderer in BigDataViewer and Paintera.

### DIFF
--- a/src/main/java/bdv/fx/viewer/render/MultiResolutionRendererFX.java
+++ b/src/main/java/bdv/fx/viewer/render/MultiResolutionRendererFX.java
@@ -36,9 +36,12 @@ import bdv.viewer.render.AccumulateProjectorFactory;
 import net.imglib2.Interval;
 import net.imglib2.Volatile;
 import net.imglib2.img.array.ArrayImg;
+import net.imglib2.img.basictypeaccess.IntAccess;
+import net.imglib2.realtransform.AffineTransform3D;
 import net.imglib2.type.numeric.ARGBType;
 import net.imglib2.ui.RenderTarget;
 import net.imglib2.ui.Renderer;
+import net.imglib2.ui.TransformListener;
 
 /**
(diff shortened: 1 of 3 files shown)


hanslovsky:
Great, thanks for looking into this @maarzt 
This will be super helpful! Should the multi-resolution renderer (eventually) live outside bdv-core? Maybe in some imglib2 package?
cc @tpietzsch @axtimwalde for opinions

igorpisarev:
Hi @maarzt, thanks for the efforts!
I've modified `MultiResolutionRendererGeneric` here a few months ago as part of #211 to allow rendering arbitrary parts of the screen, which involved changing the logic quite a bit (as far as I remember). Let me know if I can be of any help in integrating these changes into BDV.

maarzt:
#211 is actually the main motivation to look into this. @hanslovsky @igorpisarev I have some questions.
1. You never use ViewerState in Paintera's `MultiResolutionRendererGeneric` What's the motivation for that?
2. When is setScreenScales called? Is it used often?

(I'm on holidays for one week. I will work on this afterwards.)

hanslovsky:
> You never use ViewerState in Paintera's MultiResolutionRendererGeneric What's the motivation for that?

We do not use the bdv `ViewerState` in Paintera, so we cannot pass it into the multi-resolution renderer. Ideally, the (generic) multi-resolution renderer would not have any bdv dependencies but replacing all uses of bdv's `Source` was prohibitively laborous at that time.

> When is setScreenScales called? Is it used often?

This is used to adjust screen scales if rendering speed is too slow, e.g. for hidpi displays. This is usually set through the graphical user interface and thus called very rarely, e.g. once per Paintera session.

PR #269, opened 2019-07-26
Increase label render speed

axtimwalde:
Increases label render speed significantly, particularly for larger selections.  Adding and removing of objects to larger selections also became faster but could still be improved by updating the redundant segment set surgically instead of globally.


hanslovsky:
I went over the code, looks good to me. I tried with an existing project to test (de-)serialization and it works. Label rendering is faster now, I will go ahead and merge.

PR #278, opened 2019-08-01
Select All actions

igorpisarev:
* New `Ctrl`+`A` action for selecting all existing ids
* New `Ctrl`+`Shift`+`A` action for selecting all ids in the current viewer
* Do not allow to select the background label (id=0)
* Add checkbox to enable/disable mesh list tab in the side panel. Off by default because it significantly slows down bulk label operations.

All ids are collected by looping over the lowest resolution label multiset dataset (fast), or in case the labels are stored as primitive type, by looping over the full resolution dataset (can be very slow). Though, it seems that in our case labels are always stored as multisets, so it shouldn't be a problem for us.


src/main/java/org/janelia/saalfeldlab/paintera/control/IdSelector.java
```diff
@@ -53,10 +67,92 @@ public MouseClickFX appendFragmentWithMaximumCount(final String name, final Pred
...
-			final LockedSegments lock)
+	// TODO: use unique labels to collect all ids; caching
+	public void selectAll()
+	{
+		final TLongSet allIds = new TLongHashSet();
+		if (source.getDataType() instanceof LabelMultisetType)
+			selectAllLabelMultisetType(allIds);
+		else
+			selectAllPrimitiveType(allIds);
+		LOG.info("Collected {} ids", allIds.size());
```
hanslovsky: Can we change this to `LOG.debug`?

src/main/java/org/janelia/saalfeldlab/paintera/control/IdSelector.java
```diff
@@ -53,10 +67,92 @@ public MouseClickFX appendFragmentWithMaximumCount(final String name, final Pred
...
+				final long id = entry.getElement().id();
+				if (foregroundCheck.test(id))
+					allIds.add(id);
+			}
+		}
+	}
+
+	private void selectAllPrimitiveType(final TLongSet allIds)
+	{
+		LOG.warn("Label data is stored as primitive type, looping over full resolution data to collect all ids -- SLOW");
```
hanslovsky: Maybe we can make this cancellable (e.g. if the user hits escape/clicks an id etc.). This should not be part of this PR but maybe add a // TODO here

src/main/java/org/janelia/saalfeldlab/paintera/control/IdSelector.java
```diff
@@ -88,19 +186,19 @@ public void accept(final MouseEvent e)
...
 
 	private class SelectFragmentWithMaximumCount extends SelectMaximumCount
 	{
 
 		@Override
-		protected void actOn(final long id, final SelectedIds selectedIds)
+		protected void actOn(final long id)
 		{
-			if (Label.regular(id))
+			if (foregroundCheck.test(id))
```
hanslovsky: Does this mean we cannot (click-)select background anymore? I am ok with that even though sometimes the background can look cool :smile:

src/main/java/org/janelia/saalfeldlab/paintera/state/LabelSourceStateIdSelectorHandler.java
```diff
@@ -30,6 +31,8 @@
 
 	private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+	private static final LongPredicate FOREGROUND_CHECK = t -> t > 0;
```
hanslovsky: This should probably be `Label.isValid(d) && t >0`. We might have a very large valid `uint64` label that would be `< 0` if interpreted as `long`/`int64`.

src/main/java/org/janelia/saalfeldlab/paintera/state/LabelSourceStateMergeDetachHandler.java
```diff
@@ -34,24 +34,19 @@
...
 import net.imglib2.type.label.Label;
 import net.imglib2.type.numeric.IntegerType;
-import net.imglib2.view.IntervalView;
-import net.imglib2.view.Views;
 
 public class LabelSourceStateMergeDetachHandler {
 
 	private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+	private static final LongPredicate FOREGROUND_CHECK = t -> t > 0;
```
hanslovsky: This should probably be Label.isValid(d) && t >0. We might have a very large valid uint64 label that would be < 0 if interpreted as long/int64.

src/main/java/org/janelia/saalfeldlab/paintera/control/IdSelector.java
igorpisarev: Yes, from what I've seen, it happens too often that one accidentally selects the background when trying to click a small object (or clicking on its edge). We can make this a setting whether the background is selectable or not, but I can't think of any use-case where it would be needed

src/main/java/org/janelia/saalfeldlab/paintera/state/LabelSourceStateIdSelectorHandler.java
igorpisarev: Good point. I've seen similar predicates in other places in the code so maybe they would need to be changed as well.
Just now I realized that instead of defining this predicate I could've modified the `Label.regular()` method to return `false` for the background label. Though, this could've had further implications: for example, it would not allow to flood-fill the background, while currently it's allowed because it doesn't enforce the foreground test. (I assume it's a common use case to paint the boundaries of an object and then fill)

hanslovsky: I agree that the `Label.regular` method should remain as it is but we could add another method `Label.isForeground` (can be local to Paintera for now).

src/main/java/org/janelia/saalfeldlab/paintera/control/IdSelector.java
hanslovsky: Excluding foreground for this is fine with me.

src/main/java/org/janelia/saalfeldlab/paintera/state/LabelSourceStateIdSelectorHandler.java
igorpisarev: For now I changed the foreground condition in relevant places, but I agree that a method like `Label.isForeground()` would be nice to have. This would need to go into `imglib2-label-multisets`, but Paintera currently depends on a very outdated version of this package. I'll create another issue to bump the label dependencies

hanslovsky:
Looks good to me. We can think about adding an indicator similar to when persisting the canvas into background or for flood fill, including a cancellation option. But that can be addressed in a separate PR.

igorpisarev:
Sounds good! The datasets I've tested it on so far were stored as label multisets, so collection of the ids was very fast. But if it's slower on other kinds of data, I agree that the progress indicator would be helpful.

PR #284, opened 2019-08-05
Add additional group to MeshManagerWithAssignmentForSegments

### LINKED ISSUE #283: Arbitrary meshes not rendered in current master
Bisecting to find the cause.

hanslovsky:
Fixes #283

[`root.getChildren().setAll`](https://github.com/saalfeldlab/paintera/blob/master/src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java#L245) would remove all meshes that are not managed by the mesh manager. To solve this, add a separate group that is guaranteed to hold only meshes of the mesh manager.

Let me know what you think @igorpisarev


igorpisarev:
Looks good! Thanks for the fix @hanslovsky

PR #291, opened 2019-08-07
Fix deadlock at startup

### LINKED ISSUE #239: 0.14.1 breaks Lauritzen projects with synapse rendering
Starting Paintera with a Lauritzen project with synapse rendering hangs on start-up with a thread deadlock before the UI shows up. This happens with 0.14.1 but not with 0.14.0, so it may have been introduced in #222. I do not have the time to look into it right now, so I recommend users working on Lauritzen projects with syanpse rendering stick with 0.14.0 for now (you will only miss out on global interpolation instead of per source interpolation). I might have a little bit of free time at the hackathon on Wednesday and hopefully, I can fix this.

igorpisarev:
Closes #239 
The cause for the deadlock was that the list of sources in `ViewerState` was modified under the lock, and since it was an `ObservableList`, it immediately notified the listeners about the change while still holding the lock.

I've changed it to a regular list and `ViewerState` to implement `Observable` instead. It will fire the invalidated event after releasing the lock which eliminates the risk of running into a deadlock.
For consistency it will also fire when the viewer transform or the timepoint changes. I also slightly changed types and visibility of the class members in `ViewerState` to make it more straightforward.

### DIFF
--- a/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
+++ b/src/main/java/bdv/fx/viewer/ViewerPanelFX.java
@@ -186,10 +186,6 @@ public class ViewerPanelFX
 		this.renderingExecutorService = Executors.newFixedThreadPool(optional.values.getNumRenderingThreads(), new RenderThreadFactory());
 		options = optional.values;
 
-		this.state = new ViewerState(axisOrder);
-
-		state.numTimepoints.set(numTimepoints);
-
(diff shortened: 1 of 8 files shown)


hanslovsky:
Great, thanks @igorpisarev 

I will have a look soon and will merge quickly so we can have a timely patch release with this. I will also add a [FIX] tag to the commit message because I am trying to automate the release notes from merge commits.

PR #304, opened 2019-08-17
Make project directory mutable

### LINKED ISSUE #190: Make project directory mutable/save as
Currently, project directory is immutable and users are required to specify a project or use a temporary directory when starting paintera. This directory cannot be changed. While managing the project lock file is very convenient this way, this is not a great user experience. Instead, when starting Paintera without project directory argument, the project directory should be undefined and the `Save` button should open a directory choosing dialog. Additionally, a `Save As` button should be introduced that allows the user to change the project directory.
When changing (or setting) the project directory, Paintera should
 1. if directory does not exist
   - create directory
   - create project file
 2. if directory exists and contains valid paintera project
    - ask user to overwrite
    - if lock file already exists ask user if project directory should be used anyway
 3. if insufficient rights, inform user and go back to directory chooser
 4. acquire lock file
 5. release old lock file, if appropriate

As a bonus, store sources relative to project.

### LINKED ISSUE #245: Use JavaFX KeyCombination instead of KeyTracker
Currently, a `KeyTracker` keeps track of all the currently pressed keys. I am not sure that this is necessary and we should probably use JavaFX `KeyCodeCombination` instead, e.g. in a different project (Kotlin):
```kotlin
val increaseFontKeys = setOf(KeyCodeCombination(KeyCode.EQUALS, KeyCombination.CONTROL_DOWN, KeyCodeCombination.SHIFT_ANY))
val decreaseFontKeys = setOf(KeyCodeCombination(KeyCode.MINUS, KeyCombination.CONTROL_DOWN, KeyCodeCombination.SHIFT_ANY))

repl.getNode().addEventHandler(KeyEvent.KEY_PRESSED) {
    if (increaseFontKeys.any { c -> c.match(it) }) {
        it.consume()
        repl.increaseFontSize()
    } else if (decreaseFontKeys.any { c -> c.match(it) }) {
        it.consume()
        repl.decreaseFontSize()
    }
}
```

This would also make a help message a lot easier (we would just need to gather all the `KeyCodeCombinations`.

### LINKED ISSUE #256: Remove SourceInfo from Properties
That will make serialization much more straight forward, no need to add unnecessary listeners anymroe.

hanslovsky:
This is WIP with multiple targets:
 - [x] make project directory mutable
 - [x] add menu bar on top
    - [x] toggle between two different modes:
        1. [x] as top element of `BorderPane`
        2. [x] as overlay over center element of `BorderPane`
    -  [x] toggle visibility
    - [x] make all (or most) key event functionality available in menu bar:
        - [x] open data
        - [x] save project
        - [x] save project as
        - [x] toggle menu bar visibility/mode
        - [x] show paintera version
        - [x] show readme (WIP)
        - [x] all other actions
 - [x] use `KeyCombination.match` instead of `KeyTracker` for all key event handling. For mouse based events with keys, we probably still need the key tracker, unfortunately.
 - [x] add key combinations to `Properties2` class.

This will most likely be merged only after #267 

Fixes #190 
Fixes #245 
Fixes #256


hanslovsky:
I checked the side bar and all the preference nodes seem to be present after refactoring.

hanslovsky:
As far as I can tell, all key bindings that do not include any mouse event handling are now converted into `KeyCombination` style matching and are documented in the `F1` help window in the UI.

This is pretty much ready for merge but the mouse event handling needs some work, which should happen in a separate PR once this is merged.

hanslovsky:
I think I found a way to ship the readme with Paintera:
 1. convert `README.md` to `src/main/resources/README.html`, e.g. with `pandoc`
 2. `ln -s $(realpath img) src/main/resources` (relative link is better but you get the idea, can even be done permanently on GH)
 3. Inside Paintera, do something like this:

```kotlin
import javafx.application.Application
import javafx.scene.Scene
import javafx.scene.web.WebView
import javafx.stage.Stage

class WebViewIssue : Application() {
	override fun start(primaryStage: Stage) {
		val url = javaClass.getResource("README.html")
		val wv = WebView().also { it.engine.load(url.toString()) }
		primaryStage.scene = Scene(wv)
		primaryStage.show()
	}
}

fun main() = Application.launch(WebViewIssue::class.java)
```

hanslovsky:
Relevant for the rendering of resources: https://community.oracle.com/thread/2332194

hanslovsky:
README Rendering works now if classes are loaded from `.class` files, i.e. when run from Intellij IDE but images are not rendered when run from `.jar` file. This might be good enough for now but should be investigated further. I will test on my other machine and then probably merge soon.

hanslovsky:
FWIW, this is how I render the `README.md` from md into html at runtime:
```kotlin
val vs = Version.VERSION_STRING
val tag = if (vs.endsWith("SNAPSHOT"))
	"master"
else
	"^.*-SNAPSHOT-([A-Za-z0-9]+)$"
			.toRegex()
			.find(vs)
			?.let { it.groupValues[1] }
			?: "paintera-$vs"
val ghurl = "https://github.com/saalfeldlab/paintera/blob/$tag/README.md"
val jarPath = javaClass.protectionDomain.codeSource.location.let { URLDecoder.decode(it.file, "UTF-8") }
// TODO what is the correct prefix when loading from jar?
val prefix = if (jarPath.endsWith(".jar")) "jar:file:$jarPath!" else "file://$jarPath"
val extensions = listOf(TablesExtension.create())
val parser = Parser.builder().extensions(extensions).build()
val renderer = HtmlRenderer.builder().extensions(extensions).build()
val document = javaClass.getResource("/README.md").openStream().use { parser.parseReader(it.reader()) }
val readmeHtml = renderer.render(document).replace("img src=\"img", "img src=\"${prefix}/img")
val dialog = PainteraAlerts.information("_Close", true).also { it.initModality(Modality.NONE) }
val wv = WebView()
		.also { it.engine.loadContent(readmeHtml) }
		.also { it.maxHeight = Double.POSITIVE_INFINITY }
```

hanslovsky:
I will go ahead and merge. I will work with @igorpisarev to fix any merge issues with #267 once that PR is ready to merge.
I will add a new `paintera-experimental` command to saalfeldlab/paintera-cmd

PR #312, opened 2019-08-25
Consider min of source interval when centering at a source

### LINKED ISSUE #307: Ortho slices out of view
When the data source offset is not set to (0,0,0) the 3d ortho slices are not centered. This is problematic for working with datasets extracted from larger regions which tend to have very large offsets (in world coords) - the user needs to find the planes and manually center them. An immediate fix on the users end is to just set the raw offset to (0,0,0) and calculate the new offset for the labels when loading the datasets. However, it would be preferable for the ortho slices to automatically be centered regardless of the offset.

hanslovsky:
Fixes #307

Before I merge, please confirm that this fixes the issue for you @sheridana

One way to get this to run, is
```
git clone https://github.com/saalfeldlab/paintera
cd paintera
git checkout fix-307
mvn clean install
jgo -u org.janelia.saalfeldlab:paintera:0.19.3-SNAPSHOT:@Paintera+org.slf4j:slf4j-simple:1.7.25
```

If you run this from within a conda environment with paintera, `jgo` will be available. You can install `jgo` through conda
```
conda install -c conda-forge jgo
```
or pip
```
pip install jgo
```
otherwise.


hanslovsky:
Fixed after confirmation from @sheridana

PR #316, opened 2019-09-04
Remove global cache

### LINKED ISSUE #309: MaskedSource resources are not cleaned up.
After Elliot had continuing issues with unreasonably high memory consumption in his Paintera sessions, @bogovicj and I started looking into it in more detail. It appears that during those sessions, a lot of threads by name patterns of `io-sync-` and `Fetcher-` are created but do not do any work. After some digging I found out that these threads are created by the `IoSync` and `FetcherThreads` classes of imglib2-cache, respectively. These classes are used in the `DiskCachedCellImg` and `WeakRefVolatileCache`, which are used for the canvases and mask of a `MaskedSource`. Apparently, they keep running forever, even when a canvas is discarded or a mask is applied (an subsequently discarded). As the threads access (some of  the) resources of the enclosing classes, those resources can not be garbage collected. In particular, `IoSync.Writer` is an inner class and the entire `IoSync` object stays alive, as far as I understand. I confirmed by discarding the canvas every second and I saw a linear increase in live threads in visualvm. I did not confirm for the mask but it is highly likely that the same applies there.

A solution would be to free the sources by stopping those threads whenever masks or canvas are discarded. This is possible for `FetcherThreads` through the `FetcherThreads.shutdown` method but not at all for `IoSync`. Another challenge is getting access to those members in the first place. For example, this snippet gets the `IoSync` of a mask of a `MaskedSource`:
```java
try {
    if (oldDataCanvas instanceof DiskCachedCellImg<?, ?>) {
        final DiskCachedCellImg<?, ?> cached = (DiskCachedCellImg<?, ?>) oldDataCanvas;
        final Cache<?, ?> cache = cached.getCache();
        if (cache instanceof LoaderCacheAsCacheAdapter<?, ?>) {
            final LoaderCacheAsCacheAdapter<?, ?> l = (LoaderCacheAsCacheAdapter<?, ?>) cache;
            final Field field = l.getClass().getDeclaredField("loader");
            try {
                field.setAccessible(true);
                final CacheLoader<?, ?> actualLoader = (CacheLoader<?, ?>) field.get(l);
                if (actualLoader instanceof IoSync) {
                    final IoSync<?, ?> ioSync = (IoSync<?, ?>) actualLoader;
                    LOG.info("Got ioSync {}", ioSync);
                    // cannot currently stop ioSync?
                }
            } finally {
                field.setAccessible(false);
            }

        }
    }
} catch (final Exception e) {
    e.printStackTrace();
}
```
This does not include access to `FetcherThreads` of the volatile-wrapped canvas, yet. As far as I can tell, the only real solution would be functionality in imglib2-cache at some level (maybe even at the most abstract cache interface) to close/shutdown/stop a cache.

Related links:
https://gitter.im/saalfeldlab/general?at=5d5d832e7d3c163641170b8f
https://gitter.im/imglib/imglib2?at=5d5d89cb67969726f93df923

hanslovsky:
This PR removes the global cache and updates the imglib2-cache dependency to the [latest release `1.0.0-beta-12](https://github.com/imglib/imglib2-cache/releases/tag/imglib2-cache-1.0.0-beta-12) that introduces the `Invalidate` interface. The goal is to fix issues with resources that are not cleaned up when not needed any longer (#309)
 - [x] remove global cache
 - [x] replace `Invalidate` and `InvalidateAll` with imglib2-cache `Invalidate`
 - [x] Use viewer's global `SharedQueue` for canvases in `MaskedSource`
 - [x] shutdown `IoSync` when discarded. This has to wait for a release of imglib/imglib2-cache#13
 - [x] `invalidateAll` canvases when discarded
 - [x] update to non-SNAPSHOT dependency for `bigdataviewer/bigdataviewer-vistools`: We need a new release of vistools because the `SharedQueue` is not compatible with the latest imglib2-cache release.

Fixes #309


hanslovsky:
Current PR requires building with imglib/imglib2-cache#13

hanslovsky:
This should be ready to be merged but we will keep monitoring the memory usage on Elliott's before merging. Note that, currently, there is no way to invalidate the volatile canvas (only the non-volatile backing canvas) but that should be only a minor issue (if any at all) but I left a TODO comment in the code to indicate that this might need some consideration.

hanslovsky:
This should not be merged yet. I am currently observing issues when painting at low resolution: The painted mask does not get propagated into the higher resolution canvas levels.

git bisect first bad commit 87c98741202b90a6a1a63fdb25999bb525687b9d:
```
87c98741202b90a6a1a63fdb25999bb525687b9d is the first bad commit
commit 87c98741202b90a6a1a63fdb25999bb525687b9d
Author: Philipp Hanslovsky <hanslovskyp@janelia.hhmi.org>
Date:   Wed Sep 4 20:23:50 2019 -0400

    Use IoSync shutdown PR to shutdown IoSyncs
    
    Requires imglib/imglib2-cache#12

 pom.xml                                            |  2 +-
 .../saalfeldlab/paintera/data/mask/Mask.java       | 13 ++++++-
 .../paintera/data/mask/MaskedSource.java           | 44 +++++++++++-----------
 3 files changed, 36 insertions(+), 23 deletions(-)
```

```
$ git bisect log
git bisect start
# bad: [1e703b92ab4f0db998b0c6a6526265d4fddeed70] Require invalidate/volatileInvliadate/shutdown for all masks (can be nullable)
git bisect bad 1e703b92ab4f0db998b0c6a6526265d4fddeed70
# good: [b8318561679cdd328f19f1ea7ba1639d80bee81c] Remove global cache
git bisect good b8318561679cdd328f19f1ea7ba1639d80bee81c
# good: [184f86d9d05c3e8e9a1ebd3131f3a9edd80628ca] Remove obsolete class HyperEllipsoidNeighborhoodUnsafe
git bisect good 184f86d9d05c3e8e9a1ebd3131f3a9edd80628ca
# bad: [8589a47a356f29b28ce7703dd90bd27811262191] Delete obsolete class DelegateRandomAccess
git bisect bad 8589a47a356f29b28ce7703dd90bd27811262191
# good: [1d1c8d7d478606402bfdb53f43e6038f67b9313c] Invalidate data canvases when discarding
git bisect good 1d1c8d7d478606402bfdb53f43e6038f67b9313c
# bad: [87c98741202b90a6a1a63fdb25999bb525687b9d] Use IoSync shutdown PR to shutdown IoSyncs
git bisect bad 87c98741202b90a6a1a63fdb25999bb525687b9d
# good: [f700b87b0a5136f218930f7ac32f0863450432ab] Use viewer's SharedQueue for canvases in MaskedSource
git bisect good f700b87b0a5136f218930f7ac32f0863450432ab
# first bad commit: [87c98741202b90a6a1a63fdb25999bb525687b9d] Use IoSync shutdown PR to shutdown IoSyncs
```

At the moment it is unclear to me what causes this.

hanslovsky:
Mask should be invalidated/shutdown only after propagation!

https://github.com/saalfeldlab/paintera/commit/87c98741202b90a6a1a63fdb25999bb525687b9d#diff-5b950cab32344141b1cccad3c48c5381R443-R446

hanslovsky:
May be relevant:
https://stackoverflow.com/questions/31157262/javafx-garbage-collection-fails-to-clean-up-objects?rq=1

hanslovsky:
Possibly relevant:
https://stackoverflow.com/questions/31679873/why-is-this-weak-listener-not-garbage-collected

hanslovsky:
https://stackoverflow.com/questions/42690113/javafx-memory-not-freed-after-deleting-mesh

PR #319, opened 2019-09-07
Editorconfig maven plugin

hanslovsky:
Add maven plugin for [editorconfig](https://editorconfig.org). In the future, run
```
mvn editorconfig:check 
```
in continuous integration.


hanslovsky:
1. Use whatever little configuration editorconfig offers and enforce it on every maven build
 2. Anything that cannot be defined through editorconfig will fall back to IntelliJ idea default settings.

PR #324, opened 2019-09-11
copy readme instead of symlink

### LINKED ISSUE #323: Readme not available when run from jar instead of class files
The readme is not available in the help dialog (`F1`) when run from jar (i.e. in most cases) instead of from class files (e.g. when run from IDE). This is probably because `<root>/src/main/README.md` is a symlink to `<root>/README.md`. 

Fix: Remove that symlink and copy `README.md` during `validate` phase of maven:
https://stackoverflow.com/a/32081483/1725687

hanslovsky:
Fixes #323


hanslovsky:
I tried really hard to get images to work from relative paths inside the `WebView` but it just display them. May be a [long-stadning unresolved bug](https://bugs.openjdk.java.net/browse/JDK-8089600) so I will just leave it at this, for now.

Issue #332, opened 2019-09-20
Move Kotlin files into new source root at src/main/kotlin


hanslovsky:
Plugin annotation issues discovered in #331 

cf https://forum.image.sc/t/plugins-annotations-are-not-added-to-index-in-jar-file-if-implemented-in-kotlin-in-src-main-java-source-tree/29700

PR #334, opened 2019-09-22
Add src/main/java to kapt

hanslovsky:
Addresses plugin detection: Add `src/main/java` to sources for kotlin annotator processing.

@igorpisarev If you re-visit the Windows issues, please use this branch.


hanslovsky:
Temporary fix for #332 but long term, we should still move the kotlin files.

PR #340, opened 2019-10-01
Release MeshGenerator resources when disposed

### LINKED ISSUE #337: Refreshing meshes leaks memory
#316 was supposed to fix the memory issues that we observed and thought were caused by masked sources not being freed after committing and creating a new canvas (#309). However, it seems that there is still a memory leak and it seems to be related with refreshing meshes.

Ways to reproduce:
 1. Start Paintera with label data that can generate meshes
 2. Start VisualVM and connect to Paintera process
 3. Start sampler or profiler and look/filter for these classes:
     - `javafx.scene.shape.MeshView`
     - `com.sun.prism.es2.ES2MeshView`
     - `com.sun.javafx.sg.prism.NGMeshView`
 4. Press `R` key (or refresh button in the preference pane) repeatedly and monitor sampler/profiler
    - optionally, trigger garbage collection in VisualVM
 5. The number of live objects for `MeshView` will go back to the original baseline after garbage collection, but the baseline for `ES2MeshView` and `NGMeshView` continously increases

hanslovsky:
Fixes #337

Before this commit, a listener would be added to the `isManagedProperty` of the `meshSettings` to update the `MeshGenerator` with appropriate global or individual `MeshSettings`. For some reason, this kept the internal JavaFX classes `NGMeshView` and `MeshView` alive, prohibiting clean-up of their resources, and, as a result, leaking memory. This commit fixes this issue by adding a `meshSettingsProperty` to the `MeshGenerator` that the caller can `unbind` to remove the listener that holds on to the `MeshManager`'s resources.

### DIFF
--- a/src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
+++ b/src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
@@ -11,6 +11,7 @@ import javafx.beans.property.IntegerProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleIntegerProperty;
 import javafx.scene.Group;
+import javafx.scene.Node;
 import net.imglib2.FinalInterval;
 import net.imglib2.Interval;
 import net.imglib2.cache.Cache;
@@ -29,6 +30,7 @@ import org.janelia.saalfeldlab.paintera.meshes.cache.CacheUtils;
 import org.janelia.saalfeldlab.paintera.meshes.cache.SegmentMaskGenerators;
 import org.janelia.saalfeldlab.paintera.stream.AbstractHighlightingARGBStream;
 import org.janelia.saalfeldlab.util.HashWrapper;
+import org.janelia.saalfeldlab.util.NamedThreadFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -45,8 +47,10 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
@@ -82,6 +86,8 @@ public class MeshManagerWithAssignmentForSegments implements MeshManager<Long, T
 	private final List<Runnable> refreshMeshes = new ArrayList<>();
 
 	private final BooleanProperty areMeshesEnabled = new SimpleBooleanProperty(true);
+
+	private final ExecutorService bindAndUnbindService = Executors.newSingleThreadExecutor(new NamedThreadFactory("meshmanager-unbind-%d", true));
 
 	public MeshManagerWithAssignmentForSegments(
 			final DataSource<?, ?> source,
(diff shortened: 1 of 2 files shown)


hanslovsky:
Not ready for merge, needs evaluation in a real work settings. Unfortunately, [cannot convert existing PR into a draft](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107/page/3).

src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
```diff
@@ -244,7 +246,8 @@ private void removeMeshes(final Map<Long, MeshGenerator<TLongHashSet>> toBeRemov
 		root.getChildren().setAll(existingGroups);
 
 		// unbind() for each mesh here takes way too long for some reason. Do it on a separate thread to avoid app freezing.
-		new Thread(() -> toBeRemoved.values().forEach(MeshGenerator::unbind)).start();
+		// TODO this should probably not create new threads, revisit!!
+		new Thread(() -> toBeRemoved.values().forEach(mg -> mg.meshSettingsProperty().unbind())).start();
```
igorpisarev: I addressed this in #267 by simply removing this line of code. JavaFX's [bind()](https://docs.oracle.com/javase/8/javafx/api/javafx/beans/property/Property.html#bind-javafx.beans.value.ObservableValue-) is said to be implemented using weak listeners (in contrast to [addListener()](https://docs.oracle.com/javafx/2/api/javafx/beans/value/ObservableValue.html#addListener(javafx.beans.value.ChangeListener)) that uses strong references), so it seems that explicit unbinding here is not necessary. Hopefully this will not introduce more memory leaks :grimacing:

igorpisarev: Sorry, disregard my previous comment -- I didn't realize that now it would remove an actual listener and not just unbind the properties

src/main/java/org/janelia/saalfeldlab/paintera/meshes/MeshManagerWithAssignmentForSegments.java
```diff
@@ -233,21 +229,13 @@ public void removeMesh(final Long id)
...
-		toBeRemoved.values().forEach(MeshGenerator::interrupt);
-
-		neurons.entrySet().removeAll(toBeRemoved.entrySet());
-		final List<Node> existingGroups = neurons.values().stream().map(MeshGenerator::getRoot).collect(Collectors.toList());
-		root.getChildren().setAll(existingGroups);
-
-		// unbind() for each mesh here takes way too long for some reason. Do it on a separate thread to avoid app freezing.
-		// TODO this should probably not create new threads, revisit!!
-		new Thread(() -> toBeRemoved.values().forEach(mg -> mg.meshSettingsProperty().unbind())).start();
+		toBeRemoved.forEach(this::removeMesh);
```
igorpisarev: Could you please test it in the following scenario: `Ctrl+A` thousands of labels (such as in CREMI sample A), then deselect all but one label. If I recall correctly, this special case implementation for removing a set of meshes at once helped to significantly increase performance in this case, otherwise it was taking forever if removing one by one.
Sorry for the ad-hoc optimization, I should've put some note in a comment about that.

hanslovsky: I'll test it on my mahcine.

hanslovsky: I do not observe it on my laptop with sample A, but I will also try on my workstation. What scale level did you try it on on your machine?

igorpisarev: I think the scale level didn't matter so it used to be slow even at the lowest resolution. The reason was that calling `removeMesh()` for each mesh would remove the groups from the scene one by one which has quadratic complexity compared to removing them at once, and also unbinding seemed to be very slow (the separate thread was a workaround for that).
Maybe this problem is gone though since you changed how the unbinding is done -- this would be great :)

hanslovsky: Strangely, it is slow on my workstation (but not on my laptop). I reverted to something more similar to what we had in c6ea9b9c and inverted the call chain to remove code duplication (`removeMesh` now calls `removeMeshes`). I introduced a new single-threaded `ExecutorService` with daemon threads that takes care of unbinding instead of creating new threads each time. I also tried to do similar optimizations for `generateMesh` (binding on a separate thread, adding all groups at once instead of one by one), but that did not seem to have an impact.

PR #343, opened 2019-10-07
Relative data location

### LINKED ISSUE #207: It should be possible to specify sources relative to project
When copying projects that contain data, this can lead to issues with paths pointing to the old data, when the intention is to create a new, encapsuled project with copies of the data.

Possible solution:
Add methods to [`LabelBlockLooup`](https://github.com/saalfeldlab/label-utilities/blob/020e8f2c17f2f57ffddc00084464f86e6c06ceae/src/main/kotlin/org/janelia/saalfeldlab/labels/blocks/LabelBlockLookup.kt):
```
fun isRelative() : Boolean
fun setRelativeTo(path: Path)
```

Instead of using generic (in a sense that they do not know about where the data is coming from) `RawSourceState` and `LabelSourceState`, create new state classes for n5 based raw and label sources that check `isRelative()` when (de-)serializing and set the relative path accordingly.

ping @axtimwalde

hanslovsky:
Fixes #207


hanslovsky:
With this branch, datasets relative to the project directory can be loaded now, even if the project directory location is changed, e.g. download [this project](https://my.pcloud.com/publink/show?code=XZzxULkZV9AmwJpF3a0ezo85FtaUQ5TO4C9y) and open it with Paintera built from this branch.

Existing projects are still supported but cannot (currently) be updated automatically to support relative dataset locations. This can be a future milestone but is not a requirement for this feature branch, as far as I am concerned.

hanslovsky:
Two more questions remain:
 1. Should we do something similar for raw source? The case is a little different there as (currently) we do not have additional information (like id service) provided by the backend besides the `Source` and it just works already.
 2. Should we add "converting" deserializers for `LabelSourceState` that try to convert into `ConnectomicLabelState` when a project is loaded to ensure that old projects are being updated appropriately?

hanslovsky:
This PR will be ready for testing, soon. I still need to add converting de-serializers for `ChannelSourceState` and move the source creation from the backend into the individual states. I expect that to be done tomorrow (Wednesday, 2019/11/27) and will deploy for testing ASAP.

hanslovsky:
I tested with @cpatrickc and it worked for him, as well. This should be ready for a merge but I will add:
 1. a dialog to ask the user if they would like to convert the project
 2. backups of converted attributes files

hanslovsky:
Now, users can choose if they would like to convert/update deprecated datasets when deserializing a project. Affected attributes files (Paintera proejct, label data if Paintera dataset format is used, and `label-to-block-mapping` inside label Paintera datasets) are backed up in the same directory as they exist, with a timestamp suffix (`.bkp.yyyy-mm-dd_HH-mm-ss`).

While implementing this, I found a few bugs in the `ChannelSourceStateFallbackDeserializer` that I fixed. I believe that this is ready for a merge now (after fixing the merge conflict).

hanslovsky:
This should be ready for a merge now.

PR #351, opened 2019-11-18
Fix maximization of current viewer

### LINKED ISSUE #350: Maximizing current view doesn't work
It seems that in recent versions the `M` shortcut started acting the same way as `Shift`+`M`: instead of maximizing the current view it puts it side by side with the 3D view.

igorpisarev:
Fixes #350. For some reason the object bindings `maximizedColumn` and `maximizedRow` don't get updated when their observed values change. I haven't been able to figure out why the bindings don't work properly in this case and will try to extract a minimal example that reproduces this problem.
In this fix I just removed the object bindings because they're not used anywhere, and changed to return the maximized row/column values directly.


hanslovsky:
@igorpisarev I upated the master to use saalfx and the `GridConstraintsManager` now lives in that project, hence the merge conflict. I updated and released saalfx, so the solution should be as simple as bumping the saalfx dependency to version `0.1.1`. Can you try this?

igorpisarev:
Just bumped the dependency and it works properly

hanslovsky:
Thanks @igorpisarev

igorpisarev:
@hanslovsky it seems that I said it too soon and now there is another bug: `Shift+M` now doesn't work as expected and behaves the same way as `M`. Not sure if I actually tested it and it worked at first or if I missed this case, but just ran into this issue in the current master

PR #355, opened 2019-12-05
Always change brush radius by multiplication and remove upper bound for brush radius

### LINKED ISSUE #354: Brush size cannot be arbitrarily small.
At the moment, the brush size cannot be set arbitrarily small. This is a [combination of setting the increasing or decreasing the brush radius either by multiplication or by addition, depending on the current size, and its bounds, zero (good) and screen size (bad)](https://github.com/saalfeldlab/paintera/blob/3170020cd55a9a4f3327e4773117ef9f96fe61b5/src/main/java/org/janelia/saalfeldlab/paintera/control/paint/PaintActions2D.java#L238-L260):
```java
public void decreaseBrushRadius()
{
	setBrushRadius(
			Math.min(
					brushRadius.get() - brushRadiusIncrement.get(),
					brushRadius.get() / brushRadiusScale.get()));
}

public void increaseBrushRadius()
{
	setBrushRadius(
			Math.max(
					brushRadius.get() + brushRadiusIncrement.get(),
					brushRadius.get() * brushRadiusScale.get()));
}

public void setBrushRadius(final double radius)
{
	if (radius > 0 && radius < Math.min(viewer.getWidth(), viewer.getHeight()))
	{
		this.brushRadius.set(radius);
	}
}
```

I suggest the following:
 1. Always change the radius by multiplication (will go down to 0 asymptotically)
 2. Remove the upper limit (optional)
 3. Add UI element to explicitly set the radius.

hanslovsky:
Fixes #354

I removed the upper bound for the brush radius completely but it may make sense to actually have an upper bound, e.g. the longest diagonal of the dataset in world space.

### DIFF
--- a/src/main/java/org/janelia/saalfeldlab/paintera/control/paint/PaintActions2D.java
+++ b/src/main/java/org/janelia/saalfeldlab/paintera/control/paint/PaintActions2D.java
@@ -157,8 +157,7 @@ public class PaintActions2D
 			Optional.ofNullable(maskedSource.get()).ifPresent(ms -> ms.applyMask(
 					canvas.get(),
 					interval.get(),
-					FOREGROUND_CHECK
-			                                                                    ));
+					FOREGROUND_CHECK));
 		}
 
 	}
@@ -172,8 +171,6 @@ public class PaintActions2D
 	private final BrushOverlay brushOverlay;
 
 	private final SimpleDoubleProperty brushRadius = new SimpleDoubleProperty(5.0);
-
-	private final SimpleDoubleProperty brushRadiusIncrement = new SimpleDoubleProperty(1.0);
 
 	private final SimpleDoubleProperty brushRadiusScale = new SimpleDoubleProperty(1.1);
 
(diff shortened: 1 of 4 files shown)


hanslovsky:
Brush radius (`> 0`) and resize scale factor (`> 1`) can now be adjusted through the UI in the `Canvas` tab of a label source.

hanslovsky:
@axtimwalde @igorpisarev What do you think, should the brush radius have an upper bound? I removed it because it was buggy. On the other hand, it does not make any sense to paint with a larger brush than the source, so we could limit the brush radius to the longest spatial diagonal.

axtimwalde:
That's true but also very large.  Not sure what the benefit over not having an upper bound would be?

hanslovsky:
Mostly, it is unclear what the upper bound should be. The current screen size? At least in the [current (master branch) implementation](https://github.com/saalfeldlab/paintera/blob/3170020cd55a9a4f3327e4773117ef9f96fe61b5/src/main/java/org/janelia/saalfeldlab/paintera/control/paint/PaintActions2D.java#L256), this will run into issues, e.g. zooming in after setting the brush size, it could happen that now the `radius` is bigger than the viewer window, and the brush radius would not get updated. If there is a upper bound, it should probably not depend on a variable quantity like the viewer window size but on the data somehow. But I have no good idea, what this upper limit could be.

igorpisarev:
I find the current screen size to be a reasonable upper limit, since in my understanding it doesn't make sense for the user to paint outside the visible area. We could also adjust the brush size whenever the viewer window size changes. Or is there something that I'm missing?

hanslovsky:
The current upper limit is buggy (using the width and height of the viewer in viewer space but does not transform to world space). I agree that an upper limit makes sense but I do not like to implicitly change the brush size. That is something the user should have control over. Finding a good upper limit can be discussed in a separate issue though. If there are no concerns with everything else in this PR, I will go ahead and merge.

PR #360, opened 2019-12-12
Segment counts

hanslovsky:
This is a WIP draft PR to track work on segment counts and how to expose them through the UI to navigate to the (next) smallest segment.


hanslovsky:
Unfortunately, I do not think, that this will work without an explicit feature store concept. Trying to extract the counts for a larger block than CREMI (e.g. lauritzen 03) takes forever and freezes on my relatively powerful laptop. In order to proceed with this, we will need to create a fragment feature store that is pre-computed (e.g. when converting into Paintera data format) with only targeted updates after painting.

PR #367, opened 2019-12-19
Add option to disable multi-box overlay

hanslovsky:
For some screenshots it is desirable to not have the overlay


hanslovsky:
This should be ready for merge now.

PR #368, opened 2019-12-20
Fix key binding for 'select all' action

igorpisarev:
'Select all' is broken in most recent master because of the incorrect key binding.


igorpisarev:
Merged this hotfix without marking it as [BUGFIX] since the error slipped in somewhere in `relative-data-location` which hasn't been released yet.

PR #378, opened 2020-01-14
Fix display status of ConnectomicsLabelState

### LINKED ISSUE #370: Bugs in display status of ConnectomicsLabelState
It seems that a few bugs sneaked in when the display status of `LabelSourceState` was ported to `ConnectomicsLabelState`:
* Source name is not displayed
* When no label is selected, the colored square still displays the last selected color (or black if the application was just started and no label was selected yet)
* The progress indicator is not displayed for 3D fill. There is also no way of cancelling 3D fill (ESC doesn't work either now)

igorpisarev:
Fixes #370

Progress indication and cancellation of 3D flood-fill didn't work because `ConnectomicsLabelState` didn't implement `HasFloodFillState`. I added an implementation of this interface and also implemented the rest of the 'marker' interfaces similar to how it's done in `LabelSourceState` (such as `HasMeshes`, `HasSelectedIds`, and so on). I'm not quite sure where and if it's actually used, but some of these interfaces are a bit superfluous, and perhaps they can be removed or merged into one less granular interface.

### DIFF
--- a/src/main/kotlin/org/janelia/saalfeldlab/paintera/state/label/ConnectomicsLabelState.kt
+++ b/src/main/kotlin/org/janelia/saalfeldlab/paintera/state/label/ConnectomicsLabelState.kt
@@ -94,35 +94,25 @@ class ConnectomicsLabelState<D: IntegerType<D>, T>(
 	labelBlockLookup: LabelBlockLookup? = null)
 	:
 	SourceStateWithBackend<D, T>,
-	HasMeshes<TLongHashSet>,
-	HasMeshCache<TLongHashSet>,
-	HasIdService,
-	HasSelectedIds,
 	HasHighlightingStreamConverter<T>,
-	HasMaskForLabel<D>,
 	HasFragmentSegmentAssignments,
-	HasLockedSegments,
 	HasFloodFillState {
 
 	private val source: DataSource<D, T> = backend.createSource(queue, priority, name, resolution, offset)
 	override fun getDataSource(): DataSource<D, T> = source
 
 	private val maskForLabel = equalsMaskForType(source.dataType)
-	override fun maskForLabel(): LongFunction<Converter<D, BoolType>>? = maskForLabel
 
 	val fragmentSegmentAssignment = backend.fragmentSegmentAssignment
 	override fun assignment(): FragmentSegmentAssignmentState = fragmentSegmentAssignment
 
 	val lockedSegments = LockedSegmentsOnlyLocal(Consumer {})
-	override fun lockedSegments(): LockedSegmentsState = lockedSegments
 
 	val selectedIds = SelectedIds()
-	override fun selectedIds(): SelectedIds = selectedIds
 
 	val selectedSegments = SelectedSegments(selectedIds, fragmentSegmentAssignment)
 
 	private val idService = backend.createIdService(source)
-	override fun idService(): IdService = idService
 
 	private val labelBlockLookup = labelBlockLookup ?: backend.createLabelBlockLookup(source)
 
@@ -145,7 +135,6 @@ class ConnectomicsLabelState<D: IntegerType<D>, T>(
 		{ SoftRefLoaderCache<ShapeKey<TLongHashSet>, VertexNormalPair>().withLoader(it) },
 		meshManagerExecutors,
 		meshWorkersExecutors)
-	override fun meshManager(): MeshManager<Long, TLongHashSet> = meshManager
 
 	private val paintHandler = LabelSourceStatePaintHandler(selectedIds, maskForLabel as LongFunction<Converter<*, BoolType>>)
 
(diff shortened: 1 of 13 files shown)


hanslovsky:
>  I'm not quite sure where and if it's actually used, but some of these interfaces are a bit superfluous, and perhaps they can be removed or merged into one less granular interface.

This probably is still a remnant of when the viewer (and not the source) had the responsibility of creating and managing controls, UI elements, etc.

Removing this interfaces would definitely be useful but that can happen outside of this PR

src/main/kotlin/org/janelia/saalfeldlab/paintera/state/label/ConnectomicsLabelState.kt
```diff
@@ -88,27 +91,46 @@ class ConnectomicsLabelState<D: IntegerType<D>, T>(
...
+	SourceStateWithBackend<D, T>,
+	HasMeshes<TLongHashSet>,
+	HasMeshCache<TLongHashSet>,
+	HasIdService,
+	HasSelectedIds,
+	HasHighlightingStreamConverter<T>,
+	HasMaskForLabel<D>,
+	HasFragmentSegmentAssignments,
+	HasLockedSegments,
+	HasFloodFillState {
```
hanslovsky: I don't think we should implement all the obsolete marker interfaces, e.g.
```
HasMeshes
HasMeshCache
HasIdService
HasSelectedIds
HasHighlightingStreamConverter
HasMaskForLabel
HasLockedSegments
```
appear only in `LabelSourceState` or other deprecated classes.

The classes that use `HasFragmentSegmentAssignments` and `HasFloodFillState` can probably be re-written such that these interfaces become obsolete as well (but that is part of another PR/issue).

I suggest that for this PR we only add the interface that fixes the issue, which is `HasFloodFillState` as far as I understand.

hanslovsky: cf #379

igorpisarev: I've checked the usage of these interfaces in the code, and only these three are actually used in non-trivial ways:
```
HasHighlightingStreamConverter
HasFragmentSegmentAssignments
HasFloodFillState
```
I think it makes sense to implement these to ensure consistent behavior, otherwise there are small subtle bugs as far as I understand from the code (so this PR fixes a couple more bugs that weren't identified at first, such as 2d/3d flood-fill would act on fragments when it's actually supposed to act on segments).

We can remove the rest of the implemented interfaces that are not used in the code, but logically they are very similar to the used ones. For example, `HasSelectedIds` suggests that if the code needs to do something with an active selection, it can check if it's supported in the current source and ignore the request if it's not. So if we don't implement these interfaces but continue to write new code with this model in mind, we'll likely have more bugs.

@hanslovsky What do you think if we keep all implementations in this PR for now and remove them later when we revisit the interfaces and perhaps refactor the code that uses them?

hanslovsky: > We can remove the rest of the implemented interfaces that are not used in the code, but logically they are very similar to the used ones. For example, `HasSelectedIds` suggests that if the code needs to do something with an active selection, it can check if it's supported in the current source and ignore the request if it's not.

These interfaces were introduced originally to make it easier to check if a source provided a specific functionality when the viewer was still responsible for managing source-specific UI elements, controls, etc. We have now changed this and put (almost) all responsibility into each individual source. A control should never have to check for these interfaces because it can now explicitly accept all required objects in the constructor/method signature. For `HasSelectedIds`, for example, that means that code that works with `SelectedIds` always expects a `SelectedIds` to be passed as a method argument. This is in contrast of passing an arbitrary `Source` or `SourceState` that may implement `HasSelectedIds` and check for that interface within the method.

> So if we don't implement these interfaces but continue to write new code with this model in mind, we'll likely have more bugs.

On the contrary, these bugs exist because these interfaces still exist (for legacy reasons) and code has not been updated yet to explicitly pass the required objects instead of checking for the marker interfaces within the code. We should aim to remove all of these interfaces, eventually.

igorpisarev: I agree that it's much less error-prone to pass the required objects directly and the code needs to be updated to do that, but it's also important to ensure correctness in the meantime.

I've removed the unused interfaces from `ConnectomicsLabelState` (kept only the used three), and also made all of the marker interfaces deprecated.

hanslovsky: Thank you @igorpisarev

PR #380, opened 2020-01-15
Remove obsolete interfaces and classes

### LINKED ISSUE #379: Remove obsolete marker interfaces
There are a few interfaces that were used to indicate certain functionality that are still a remnant of when the viewer (and not the source) had the responsibility of creating and managing controls, UI elements, etc. Moving this responsibility into the sources made the sources more flexible and rendered those interfaces complete. We should remove those interfaces to avoid any further confusion.

See also #378

hanslovsky:
This PR is a draft to
 - [ ] remove obsolete marker interfaces
   - [x] `HasHighlightingStreamConverter`
   - [x] `HasIdService`
   - [x] `HasLockedSegments`
   - [x] `HasMaskForLabel`
   - [x] `HasMeshCache`
   - [x] `HasMeshes`
   - [x] `HasSelectedIds`
   - [x] `HasFloodFillState`
   - [x] `HasFragmentSegmentAssignments`
   - [x] `HasAxisOrder`
 - [ ] remove obsolete classes
   - [x] `Paintera` and helper classes for UI
   - [x] `AxisOrder`
   - [ ] rename replacement classes appropriately, e.g. `Paintera2` -> `Paintera`
 - [ ] move all kotlin files into `src/main/kotlin`

In the course of this clean-up, all handlers should be modified such that they use a specific source and do not request the current source from the viewer. The latter was necessary because all event handling and UI elements were managed by the viewer but with moving this responsibility into the individual sources, we can get rid of these hard-to-understand lines of code that query the viewer for the current source and check for many different interfaces.

This is a draft PR and it will probably take a while to update everything.

Fixes #379 and beyond


hanslovsky:
Closed in favor of #390

PR #382, opened 2020-01-22
Orthoslices transparency

igorpisarev:
Adds new slider in OrthoViews settings for controlling transparency of the orthoslices.

Changing the opacity property of `MeshView` didn't have any effect on transparency (only hid the orthoslices when set to 0, otherwise fully opaque), but changing the opacity of the diffuse color in `PhongMaterial` did the trick.

### DIFF
--- a/src/main/java/org/janelia/saalfeldlab/paintera/viewer3d/OrthoSliceFX.java
+++ b/src/main/java/org/janelia/saalfeldlab/paintera/viewer3d/OrthoSliceFX.java
@@ -11,7 +11,9 @@ import org.janelia.saalfeldlab.util.fx.Transforms;
 import bdv.fx.viewer.ViewerPanelFX;
 import bdv.fx.viewer.render.RenderUnit;
 import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.DoubleProperty;
 import javafx.beans.property.SimpleBooleanProperty;
+import javafx.beans.property.SimpleDoubleProperty;
 import javafx.collections.FXCollections;
 import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
@@ -70,6 +72,13 @@ public class OrthoSliceFX
 					InvokeOnJavaFXApplicationThread.invoke(() -> this.getScene().getChildren().remove(meshesGroup));
 				}
 			}
+		});
+	}
+
+	private final DoubleProperty opacity = new SimpleDoubleProperty(1.0);
+	{
+		this.opacity.addListener((obs, oldv, newv) -> {
+			((PhongMaterial) this.meshViews.get(0).getMaterial()).setDiffuseColor(new Color(0, 0, 0, newv.doubleValue()));
 		});
 	}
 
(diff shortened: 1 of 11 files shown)


hanslovsky:
Cool, I will have a look tomorrow. Thank you @igorpisarev

src/main/java/org/janelia/saalfeldlab/paintera/viewer3d/OrthoSliceFX.java
```diff
@@ -73,6 +77,20 @@
 		});
 	}
 
+	private final DoubleProperty opacity = new SimpleDoubleProperty(1.0);
+	{
+		this.opacity.addListener((obs, oldv, newv) -> {
+			((PhongMaterial) this.meshView.get().getMaterial()).setDiffuseColor(new Color(
+					textureDiffuseOpaqueColor.getRed(),
```
hanslovsky: Could we just use [`Color.deriveColor`](https://docs.oracle.com/javafx/2/api/javafx/scene/paint/Color.html#deriveColor(double,%20double,%20double,%20double)) here?
```java
textureDiffuseOpaqueColor.deriveColor(0.0, 1.0, 1.0, newv.doubleValue());
```

src/main/java/org/janelia/saalfeldlab/paintera/viewer3d/OrthoSliceFX.java
```diff
@@ -200,12 +230,16 @@ private void initializeMeshes()
...
 		mv.setCullFace(CullFace.NONE);
 		mv.setMaterial(material);
-		material.setDiffuseColor(Color.BLACK);
-		material.setSpecularColor(Color.BLACK);
 
-		newMeshViews.add(mv);
+		// NOTE: the opacity property of the MeshView object does not have any effect.
+		// But the transparency can still be controlled by modifying the opacity value of the diffuse color.
+		material.setDiffuseColor(new Color(
+				textureDiffuseOpaqueColor.getRed(),
```
hanslovsky: Could we just use [`Color.deriveColor`](https://docs.oracle.com/javafx/2/api/javafx/scene/paint/Color.html#deriveColor(double,%20double,%20double,%20double)) here?
```java
textureDiffuseOpaqueColor.deriveColor(0.0, 1.0, 1.0, newv.doubleValue());
```

igorpisarev:
Thanks @hanslovsky, actually I'm still working on this PR to resolve the remaining transparency issue and make the intersection lines more visible. I'll update the PR once it's done.

src/main/java/org/janelia/saalfeldlab/paintera/viewer3d/OrthoSliceFX.java
igorpisarev: I didn't know about this method, I think I'll switch to it as you suggested to make it more concise

hanslovsky:
Great, that works for me.

igorpisarev:
Transparent orthoslices are finally working:

transparent-orthoslices

Each orthoslice is now composed of 4 mesh views (each quarter of an orthoslice is a separate `MeshView` consisting of two triangles), and all these mesh views are added to the same display group. JavaFX renders the objects in a group from first to last, so whenever the transform changes, the mesh views are re-ordered based on their distance to the camera such that farthest meshes come first.

I'll clean up the code a bit more and this will be ready for merging.

igorpisarev:
I've finally fixed the remaining issue with transparency. Previously it was "sort of" working most of the time, but as it turned out there were cases when under certain angles the order wasn't correct. (I was sorting the quadrants of each orthoslice by the z-plane of its center point, but this approach sometimes gives incorrect results with perspective camera).

The new and proper way of doing this is to determine the side of each orthoslice that we're currently seeing (front or back), and then based on that establish the ordering of the quadrants. There are 8 possible orientations of front/back sides for 3 orthoslices, and each of them allows to split the quadrants into three groups (closest, farthest, in-between).

#### Before / after:
orthoslices-comparison

---------------------

There is also now a new slider to control shading of the orthoslices. Basically this allows to tweak the diffuse component, where the old behavior is 0 (no diffuse component, only the self-illumination map). Adding the diffuse component makes it easier to see the intersections more clearly.

This is now ready for merging.

Issue #376, opened 2020-01-14
Paint expansion


dnguyen-thi:
Hi,

I wanted to mention an issue that I have been experiencing while working with Paintera.
We are currently using Paintera to achieve dense segmentation in our dataset.

First, the smallest brush size is still bigger than a voxel. In fact, it usually paints in a (+) shape which makes reaching certain corners difficult.
Additionally, the paint tends to expand after Paintera computes the changes I make. For instance, if I intended to paint 5 voxels, 7-8 voxels end up with that segment ID. 

Do you have a proposed solution for this issue I am experiencing? 

Thanks!

hanslovsky:
Thank you for reporting this issue @dnguyen-thi 

I will start with (part of) your first issue, the smallest brush size being bigger than a voxel: This was reported in #354 and fixed in #355. We are trying to release Paintera this week or next week so, hopefully, this will be fixed automatically, soon.

Regarding the expanding paint, I do not have an answer at this moment and we will have to investigate, what exactly is going wrong there. I tried for myself and I can confirm this erroneous behavior of Paintera. It seems that the resolution of the mask (technical detail) may be inconsistent with the resolution of the data.

cc @igorpisarev

dnguyen-thi:
Hi again!

I've recently updated Paintera to 0.22.0 and would like to mention that although the newer version does fulfill my initial request for a smaller brush size, paint expansion still occurs. For that reason, I am still experiencing the same painting restrictions as per prior to the update. 

Thank you!

hanslovsky:
Thank you for reporting back. The Paint expansion issue has not been addressed yet.

hanslovsky:
@igorpisarev I am currently looking into this and right now I have no clue where it is going wrong. I am looking at the `MaskedSource` to see if maybe the transforms are incorrect.

One thing that I noticed is that the transforms for the interpolated and scaled masks may be set incorrectly in [`MaskedSource.setMaskScaleLevels`](https://github.com/saalfeldlab/paintera/blob/074c2279da9c80844961b3e555e7dbfea9706404/src/main/java/org/janelia/saalfeldlab/paintera/data/mask/MaskedSource.java#L1544).
[...]

PR #385, opened 2020-02-10
Fix paint expansion issues

hanslovsky:
This branch/PR tries to find and fix the issue that causes #376 

Copied status report from https://github.com/saalfeldlab/paintera/pull/385#issuecomment-584892341:
I fixed the following issues:
 1. It was impossible to paint a single (data) voxel because the radius in data space was determined by ceiling the transformed radius, i.e. even if the transformed radius is small like `1e-10`, it would still end up being `1` instead of `0`, resulting in a `+`-shape. Fixed in 698e1013cdac6b126f1855cd713e4edd18d71f27
 2. Roi to be repainted in viewer was based on brush radius instead of actually painted voxels but because voxel have an extent, this may be incorrect, causing the "paint expansion" described in #376. Fixed in 63b546d820449e018cea1077a2aa3b4784dc39c8. When not axis-aligned, however, the repainted area is too small. While doing this, I also ensured that all viewers get updated, if appropriate, not just the one that is being painted in.

Additionally, I use optimized painting only if viewer is axis aligned with the data and the data is isotropic within the painting plane (550aa50cab8985b956a2e02746bd767b2cfccc6e). As a result, we do not need `HyperEllipsoidNeighborhood` and related classes anymore. Other than that, I decided to stick with the existing painting method because the small-radii issue in https://github.com/saalfeldlab/paintera/pull/385#issuecomment-584333693 is relevant only if the radius is in the order of the voxel size and I think the current behavior is ok for such corner cases.


hanslovsky:
New findings in https://github.com/saalfeldlab/paintera/issues/376#issuecomment-584287460

hanslovsky:
I am currently trying to 
 1. use the correct radii, and
 2. avoid the use of `HyperEllipsoidNeighborhood`.

(2) is simple: Use optimized painting, i.e. painting with neighborhoods in label space, only when axis-aligned **and** isotropic within the plane. However, there are some issues:
 - Some voxels are missed out sometimes (may be a bug with translating the center point and radius from world or viewer space to label coordinate space)
 - For small radii in label coordinate space, e.g. `1`, there will always be cases for which some voxel will not be painted even though they intersect with the brush overlay, e.g.
paintera-missing-intersecting-voxels
Maybe the optimized scheme should be abandoned alltogether.

There is, however, a performance issue: [Filling a transformed cylinder](https://github.com/saalfeldlab/paintera/blob/b975b81cfbfd5a073ca85bf48b0d53c223f9a809/src/main/java/org/janelia/saalfeldlab/paintera/control/paint/Paint2D.java#L97) does not seem to get the boundaries right and iterating over screen voxels is not efficient enough. This will require some more work.

hanslovsky:
I fixed the following issues:
 1. It was impossible to paint a single (data) voxel because the radius in data space was determined by ceiling the transformed radius, i.e. even if the transformed radius is small like `1e-10`, it would still end up being `1` instead of `0`, resulting in a `+`-shape. Fixed in 698e1013cdac6b126f1855cd713e4edd18d71f27
 2. Roi to be repainted in viewer was based on brush radius instead of actually painted voxels but because voxel have an extent, this may be incorrect, causing the "paint expansion" described in #376. Fixed in 63b546d820449e018cea1077a2aa3b4784dc39c8. When not axis-aligned, however, the repainted area is too small. While doing this, I also ensured that all viewers get updated, if appropriate, not just the one that is being painted in.

Additionally, I use optimized painting only if viewer is axis aligned with the data and the data is isotropic within the painting plane (550aa50cab8985b956a2e02746bd767b2cfccc6e). As a result, we do not need `HyperEllipsoidNeighborhood` and related classes anymore. Other than that, I decided to stick with the existing painting method because the small-radii issue in https://github.com/saalfeldlab/paintera/pull/385#issuecomment-584333693 is relevant only if the radius is in the order of the voxel size and I think the current behavior is ok for such corner cases.

hanslovsky:
@cpatrickc Please test this pull request whenever you have a minute and let me know what you think of the painting experience, in particular:
 1. Painting in high-resolution isotropic axis aligned both zoomed in/out
 2. Painting in low resolution anisotropic both zoomed in/out
 3. Painting non-axis aligned.

Let me know if you need help setting up your Paintera to use this branch.


Thank you!!

igorpisarev:
This works great now! I tested different scenarios and everything worked as expected, and code looks good to me too.

hanslovsky:
Thanks a lot @igorpisarev!

cpatrickc:
I tested all scenarios outlined above. The brush works great!

hanslovsky:
Great, thanks @cpatrickc

Issue #386, opened 2020-02-12
Set log level at run-time


hanslovsky:
At the moment, it is impossible to set the log level at run-time. Paintera depends on `slf4j-api` and not on a specific implementation and the `slf4j-api` does not expose a way of setting the log level at runtime. Currently, a slf4j implementation has to be added to the classpath at runtime, `slf4j-simple` in the [paintera-cmd](https://github.com/saalfeldlab/paintera) and
the log level can be set only at start-up by setting appropriate JVM environment through [-Dorg.slf4j.simpleLogger....](http://www.slf4j.org/api/org/slf4j/impl/SimpleLogger.html). The paintera project is not intended to be a library but an application (core features can be extracted for re-use, e.g. [`saalfx`](https://saalfeldlab/saalfx), and it would thus make sense to add an explicit to an `slf4j` implementation. Of all the `slf4j` implementations, only [logback](http://logback.qos.ch/) exposes API to change the log level programatically, to the best of my knowledge.

I suggest:
 1. Use explicit `sl4fj-logback` dependency
 2. Add command line argument `--log-level` to set log level for root logger/specific classes
 3. Add UI elements to control log level
 4. Remove `slf4j-simple` from paintera-cmd
 5. (Optional) If possible with logback: Add option to specifiy log target, i.e. stdout, file, or both.

hanslovsky:
Fixed in #387

PR #387, opened 2020-02-12
Use logback to configure log level at runtime

hanslovsky:
This is a draft PR to address #386 

 - [x] Use explicit `sl4fj-logback` dependency (need to exclude `slf4j-simple` because of fxyz3d dependency)
 - [x] Add command line argument `--log-level` to set log level for 
   - [x] root logger
   - [x] named loggers
 - [x] Add UI elements to control log level
   - [x] root logger
   - [x] named loggers
 - [x] Log into
   - [x] console
   -  ~UI~
   - [x] file in project directory
 - [x] Remove `slf4j-simple` from paintera-cmd https://github.com/saalfeldlab/paintera-cmd/pull/3
 - ~(Optional) If possible with logback: Add option to specifiy log target, i.e. stdout, file, or both.~

The command line interface now has options to set the root logger level
```
paintera --log-level <LEVEL>
```
and individual logger levels by name:
```
paintera [--log-level-for NAME=LEVEL[,NAME=LEVEL]...]...
paintera [--log-level-for=NAME=LEVEL[,NAME=LEVEL]...]...
```

Logger Config in the UI:


hanslovsky:
Examples for setting log directory at runtime:
https://mkyong.com/logging/logback-xml-example/
https://mkyong.com/logging/logback-set-log-file-name-programmatically/

It is unclear, though, if it is possible to update the log file once it is set.

hanslovsky:
This may be a solution:
https://stackoverflow.com/a/3810936/1725687

hanslovsky:
I think that the crossed-out bullets are not really beneficial enough to justify the effort or can be added later, if desired. This is ready for review now.

hanslovsky:
Comment from zoom meeting: 
 - Write logs into home directory instead of paintera project.
 - Add check box to toggle logging to file

hanslovsky:
@axtimwalde @igorpisarev I was able to come up with a good solution, I think: The basename of the is set as a static variable once at startup and the logging service is refreshed afterwards to use this basename:
If a process id (`PID`) is available, `BASENAME` is
```
paintera.${PID}.yyyy-MM-dd_HH-mm-ss
```
else it is
```
paintera.yyyy-MM-dd_HH-mm-ss
```
The log file is then located at
```
~/.paintera/logs/${BASENAME}.log
```

I think, this is reasonable but if you have any concerns about the exact basename, it is going to be easy to adjust. I will also show the log file location in the UI so users do not have to play a guessing game.

Let me know what you think.

igorpisarev:
@hanslovsky That's a really useful feature! I gave it a try and everything seems to work well. There is a couple of small things that I noticed:
* When application is started with an existing project, the log level of the ROOT logger is always set to `INFO` in the UI, even if it was set to something else, for example, `WARN`. It seems that it does use the correct level for logging though (`WARN` in this case), but just doesn't display the actual value in the UI.
* The package names seem to be compressed to just one letter in the logs, for example:
  ```
  2020-03-19 16:26:16,856 5634 [painter-thread] DEBUG b.f.v.r.MultiResolutionRendererGeneric - Creating projector. Pre-multiply? true
  ```
  Do you think it's possible to change it to write out full package names, such as `bdv.fx.viewer.render.MultiResolutionRendererGeneric`? I think it would make it a bit easier to locate the class in the codebase, if needed.
* I like the log filename pattern, but currently it's not always obvious which file is the most recent one. Maybe we could put PID after the date/time so that it's easy to sort the filenames in chronological order?

hanslovsky:
Thanks for testing and your suggestions @igorpisarev

>  the log level of the ROOT logger is always set to INFO in the UI

I will investigate

> Do you think it's possible to change it to write out full package names

Yes it should be, I will check how to do that with logback

>  Maybe we could put PID after the date/time so that it's easy to sort the filenames in chronological order?

Yes, I can do that, easy fix.

hanslovsky:
Fixed the abbreviated logger names in 1dd2cf8. It was as imple as changing
```
%logger{35}
```
to
```
%logger
```

hanslovsky:
I changed log file pattern to list date/time before PID in be73d1b That way it is easier to sort log files chronologically. Append with underscore instead of dot because it is more clear that it is not a fraction of a second that way

hanslovsky:
The remaining issue with the incorrect log level for the ROOT logger in the UI will probably require a little more work and I will address that tomorrow.

hanslovsky:
I think I found the reason for the ROOT logger level issue:
 1. If [`painteraArgs.logLevel != null`](https://github.com/saalfeldlab/paintera/pull/387/files#diff-aaf4f1917f1bb61872b346cc66da05d5R60), the log level is updated. If the user does not specify a log level on the command line, `painteraArgs.logLevel === null` is expected, but
 2. the [`--log-level` flag has a default value](https://github.com/saalfeldlab/paintera/pull/387/files#diff-9f5f8dae896bc1cedd0054b1faf5d2f1R436) and is never `null`, i.e. the serialized value will always be overriden by the command line value, even if the user did not specify any.

cc @igorpisarev

hanslovsky:
I will resolve the conflicting files and then merge this PR

hanslovsky:
I am not sure why GitHub says this PR has conflits. When I merge into master on the cl everything works without any interaction.

hanslovsky:
Apparently, GitHub and git evaluate merge conflicts differently:

https://github.community/t5/How-to-use-Git-and-GitHub/Merge-conflicts-on-GitHub-not-locally/td-p/7184
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-on-github

PR #394, opened 2020-02-24
Fix memory leak when meshes are refreshed

### LINKED ISSUE #393: Refreshing meshes leaks memory again
This is similar to #337 and is occurring again in the most recent release. It wasn't the case in the previous major release 0.22.0 so it must have been introduced somewhere in the process of revamping the mesh managers. I'm currently doing `git-bisect` to narrow it down.

igorpisarev:
Closes #393


igorpisarev:
I was able to reliably reproduce the memory leak on my machine and confirm that it's fixed by this change, so I'll go ahead and merge & release a new version.

PR #409, opened 2020-05-05
Add ability to export meshes for intersecting sources (synapses)

igorpisarev:
Adds new `Export` button under the `Meshes` tab in the menu of an intersecting source.

After selecting the output directory, the meshes will be stored in a file `synapses{segment_ids}.obj` (or `.bin`), so that the filename will include the current selection of segments that generate the synapse meshes of the intersecting source.


igorpisarev:
* Fixed `filename too long` error when trying to export synapse meshes for a neuron with a large number of merges. Instead of putting the full fragment ID list in the filename, we now use the segment ID list.
* A few other minor fixes related to mesh exporter dialog.

PR #454, opened 2022-06-10
Readme docs update, re-order usage table info

GenevieveBuckley:
This PR is mostly a re-ordering of existing information. Grouping the information in the "Usage" table into subsections according to function makes it easier to find the information you're looking for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,51 +268,69 @@ Usage: Paintera [--add-n5-container=<container>...
 ## Usage
 
 | Action | Description |
-| --------------- | ----------- |
-| `P`               | Show Status bar on right side |
-| (`Shift` +) `Ctrl` + `Tab` | Cycle current source forward (backward) |
-| `Ctrl` + `O` | Show open dataset dialog |
+| -------------- | ----------- |
+| **Opening data:**  |
+| `Ctrl` + `O`   | Show open dataset dialog |
+| `Ctrl` + `Shift` + `N` | Create new label dataset |
+| `P`            | Show status menu on right side |
+| `V` | Toggle visibility of current source dataset |
+| `Ctrl` + `Tab` | Cycle current source dataset forward |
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
Oh, hang on - there are two files called README.md (one in the top level of the repository, and one under `markdown/README.md`). 

Most of the content seems to be duplicated. Should I also copy my changes into the second readme file as well? What's the difference between the two?

cmhulbert:
The nested README.md should just be a symlink to the top level one, so you shouldn't need to explicitly copy the changes. Thanks for clearing up some of the README.md!

GenevieveBuckley:
> The nested README.md should just be a symlink to the top level one, so you shouldn't need to explicitly copy the changes. Thanks for clearing up some of the README.md!

Ah, of course. I probably should have looked more closely.

I'm still not entirely sure where the content is being pulled for the key bindings dialog that pops up if you press F1 (or access it from the help menu). It seems separate from the README content. I found [KeyBindingsDialog.kt](https://github.com/saalfeldlab/paintera/blob/6535fd3ea05a97d25405cceef604f78696de53fb/src/main/kotlin/org/janelia/saalfeldlab/paintera/ui/dialogs/KeyBindingsDialog.kt), but I still don't see where it's getting the content from. Ideally these tables would also be organised in the same way, but I guess that can be done in another PR.

GenevieveBuckley:
FYI, I've also added links to the two new youtube tutorial videos under the "Usage" heading. I think that's a good place for it.

I don't plan to make more changes to this PR, unless you request them.

GenevieveBuckley:
Added some more information from Stephan's [comment here](https://github.com/saalfeldlab/paintera/issues/456#issuecomment-1183363364), plus a link to the [BigDataViewer docs on navigation controls](https://imagej.net/plugins/bdv/#basic-navigation) since he says they are the same

PR #523, opened 2023-12-28
Move Running Paintera section in README

hanslovsky:
I believe the "Running Paintera" subsection needs to go under "Running from source", not "Development Releases".


cmhulbert:
Looks good, I agree, thanks!

